PR #44, opened 2018-10-02
support modifying the resource download request

OriHoch:
Added an IXLoader interface method: modify_download_request, which can be used to modify the download url or http headers.

### DIFF
--- a/ckanext/xloader/interfaces.py
+++ b/ckanext/xloader/interfaces.py
@@ -10,6 +10,19 @@ class IXloader(Interface):
     The before_submit function, when implemented
     """
 
+    def modify_download_request(self, url, resource, api_key, headers):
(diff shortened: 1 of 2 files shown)


ckanext/xloader/interfaces.py
```diff
@@ -10,6 +10,19 @@ class IXloader(Interface):
     The before_submit function, when implemented
     """
 
+    def modify_download_request(self, url, resource, api_key, headers):
+        """ Can be used to modify the http download request.
+        The headers parameter is a dict which should be modified directly.
```
ThrawnCA, line 15: I think this should say "can be modified" rather than "should be modified".

ThrawnCA:
@OriHoch This looks like it could proceed, but needs the merge conflict resolved.

Issue #45, opened 2018-10-04
Jobs are queued but not processed


stefina:
We are experiencing that our resources are not loaded into the datastore anymore. It still works fine on our integration-environment.

Executing the xloader-command to list its status like so:
```paster xloader status -c /var/www/ckan/default/development.ini```

gives me a list of jobs which are dated about a month ago until today. I was not able yet to find out how this happened but was wondering if there is a way to "flush the queue" somehow, as a reload of a resource into the datastore does not work either.

Have you experienced something similar? Maybe you can point me towards a possible way to clean up the queue or how to debug this?

stefina:
I was not able to figure out why the worker queue was not processed anymore. I figured though, that the queued jobs from about a month ago wouldn't be processed anytime soon, so I decided to clear said queue to get the xloader load our resources into the DataStore again.

If anyone stumbles into the same problem, I troubleshooted my way out of it by clearing our xloader-queue via the ckan-jobs-command:

`paster --plugin=ckan jobs clear [QUEUES] -c /var/www/ckan/default/development.ini`

You just have to be aware that everything that was queued won't be processed and you have to consider a resubmit of your jobs :P

davidread:
@stefina sorry for the delay. I'm pleased to see you found this CLI command. Some of the CLI is in xloader itself and some of the queue stuff is part of core ckan, like you found, which is confusing.

If the worker stops processing then I guess you also tried restarting it e.g.
> sudo supervisorctl restart ckan-worker:ckan-worker-00

And if that still doesn't pick up jobs I suggest running the worker on the command-line and you can even put in a pdb breakpoint and step through until it gets stuck.

stefina:
Thank you @davidread for following up on this and no worries for the delay!

Yeah, I tried restarting it which wasn't any luck. But thank you for the hint on debugging it on the command-line, that I haven't thought of yet. Will keep that in mind for next time if it happens again, which it hopefully won't :)

PR #48, opened 2018-10-05
Add basic CLI instructions

davidread:


davidread:
TODO: add info from: https://github.com/ckan/ckanext-xloader/issues/45

PR #49, opened 2018-10-10
convert values to string before encoding

stefina:
When column-headers contain floats the encoding fails as floats do not have an encode-function.

Further explanation: When a file with floats in one of the header-columns, the import failed with the error message `AttributeError: 'float' object has no attribute 'encode'`
Not sure, if my approach either has side-effects or if this is the correct approach but my import with specified file works fine after this change.

### DIFF
--- a/ckanext/xloader/loader.py
+++ b/ckanext/xloader/loader.py
@@ -63,7 +63,7 @@ def load_csv(csv_filepath, resource_id, mimetype='text/csv', logger=None):
         header_offset, headers = messytables.headers_guess(row_set.sample)
 
     # Some headers might have been converted from strings to floats and such.
-    headers = [unidecode(header) for header in headers]
+    headers = encode_headers(headers)
 
     # Guess the delimiter used in the file
     with open(csv_filepath, 'r') as f:
@@ -292,7 +292,7 @@ def load_table(table_filepath, resource_id, mimetype='text/csv', logger=None):
                 for f in existing.get('fields', []) if 'info' in f)
 
         # Some headers might have been converted from strings to floats and such.
-        headers = [unidecode(header) for header in headers]
+        headers = encode_headers(headers)
 
         row_set.register_processor(messytables.headers_processor(headers))
         row_set.register_processor(messytables.offset_processor(offset + 1))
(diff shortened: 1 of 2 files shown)


davidread:
Looks great - thanks! Do add a test for encode_headers, please.

stefina:
Thank you for the quick feedback! I added a test for the `encode_headers`-function. Just let me know if you see an issue with it!

davidread:
Is this now covered by #52 ? I couldn't reproduce the problem with this test: https://github.com/ckan/ckanext-xloader/compare/stefina-header-numbers as messytables.headers_guess() returns strings for me. (I've tried messytables 0.15.2 and the latest master)

stefina:
Ah, so true! I tested this with your branch from #52 so no worries about this PR. 

Closed and thank you :)

stefina:
I have to re-open this issue as I missed a test-case which is still not working with current master.

The test-case is quite similar: The actual issue this time is `Load using COPY failed: Error during the load into PostgreSQL: invalid byte sequence for encoding "UTF8": 0x00` (Unfortunately I haven't found a simple example for that error, yet.)
Then it tries to log floats that can not be encoded.

Not sure if my solution is the way to go in that case, but I was able to get my data imported with messytables as fallback with that fix.

davidread:
I've merged most of this in here: https://github.com/ckan/ckanext-xloader/pull/53/commits/2385cece92182699e536d34fab7bdf71bb329ba6 

Sorry I made a mistake and lost your `test_invalid_byte_sequence` test in the process. I've put it on a fresh branch here: https://github.com/ckan/ckanext-xloader/compare/test_invalid_byte_sequence and you're welcome to open a PR on it. The test fails with:
```
  File "/vagrant/src/ckanext-xloader/ckanext/xloader/tests/test_jobs.py", line 57, in wrapper
    return_value = func(*args, **kwargs)
  File "<string>", line 3, in wrapper
  File "/vagrant/src/ckanext-xloader/ckanext/xloader/tests/test_jobs.py", line 290, in test_invalid_byte_sequence
    assert result is None, jobs_db.get_job(job_id)['error']['message']
AssertionError: No entries found - nothing to load
```
which looks to me like messytables going wrong for some reason, and needs more work to fix I think.

PR #29, opened 2018-03-02
Support unicode headers in loader

metaodi:
Related to #28 and #14

PR #50, opened 2018-10-11
Deal with UniDecodeError

stefina:
When uploading a csv produces an error with `Error during the load into PostgreSQL` and the error-result contains objects that can not be decoded, a UniDecodeError is thrown and prevents the DataPusher-Mechanism from stepping in. 

Is there are cleaner and/or more pythonic ways to solve this? I'd be very happy to get some feedback even though this only fixes the outdated datapusher-logic.

### DIFF
--- a/ckanext/xloader/jobs.py
+++ b/ckanext/xloader/jobs.py
@@ -479,11 +479,11 @@ class StoringHandler(logging.Handler):
         try:
             # Turn strings into unicode to stop SQLAlchemy
             # "Unicode type received non-unicode bind param value" warnings.
-            message = unicode(record.getMessage())
-            level = unicode(record.levelname)
-            module = unicode(record.module)
-            funcName = unicode(record.funcName)
-
+            message = self._safe_encode(record.getMessage())
+            level = self._safe_encode(record.levelname)
+            module = self._safe_encode(record.module)
+            funcName = self._safe_encode(record.funcName)
+            
             conn.execute(db.LOGS_TABLE.insert().values(
                 job_id=self.task_id,
                 timestamp=datetime.datetime.now(),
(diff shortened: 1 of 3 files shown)


davidread:
@stefina thanks for this :)

I've added another test for the loader change: https://github.com/ckan/ckanext-xloader/compare/stefina-deal-with-UniDecodeError#diff-ea4b03f349ac202bb8b3f0de215e2156R230
and when I removed the loader change: https://github.com/ckan/ckanext-xloader/commit/09132f61bc54cd73bbf8f5576ddf104145ad6578#diff-a02fcaf940b7255152e941bd327be9b5R399 the test doesn't fail, as it should do. I guess my test csv doesn't have the same properties as the csv you were fixing this for? Can you help me reproduce the issue in this test that this fix is for, please?

stefina:
@davidread thank you so much for adding a test and your feedback!

A minimal example that covers this looks like this:

```
name,withumlautstrings
sophie,sophié
bernard,bernård
paul,paül, another field
```

Initially it fails because of an issue that is described here. https://github.com/ckan/ckanext-xloader/issues/51 
But as it contains Umlauts, too, it fails at this particular point and prevents the second try with messytables.

Does this help?

davidread:
It seems we already deal with umlauts on their own ok - see https://github.com/ckan/ckanext-xloader/pull/29. The problem here is when they are in a file which *also* has extra commas (noted in issue #51). I tackled the encoding issue at root in a new PR: #52 and which provides a test that shows the problem - I hope you don't mind? Thanks for flagging this up! Robustness is so important for this.

stefina:
No problem, thank you for taking time to fix it properly!

PR #52, opened 2018-10-12
Fix for umlaut AND extra comma

davidread:
A CSV with both these difficulties was causing an exception that was badly handled.

I added a test `test_umlaut_and_extra_comma` for this situation and made sure the encoding was properly handled. Now it reverts to using messytables to load it fine.


davidread:
@stefina would you mind trying this out, please?

stefina:
It works for my cases! This is great, thank you @davidread !

PR #53, opened 2018-10-19
Fix for when XLSX header contains integers

davidread:
@stefina provided an XLSX with integers in the header, which caused load_csv to fail with this exception:
```
Traceback (most recent call last):
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/vagrant/src/ckanext-xloader/ckanext/xloader/tests/test_loader.py", line 238, in test_integer_header_xlsx
    mimetype='CSV', logger=PrintLogger())
  File "/vagrant/src/ckanext-xloader/ckanext/xloader/loader.py", line 66, in load_csv
    headers = [unidecode(header) for header in headers]
  File "/usr/lib/ckan/default/local/lib/python2.7/site-packages/unidecode/__init__.py", line 48, in unidecode_expect_ascii
    bytestring = string.encode('ASCII')
AttributeError: 'float' object has no attribute 'encode'
```
and the exception was not dealt with well, so it didn't try to use messytables to load it instead. This fixes that error.

### DIFF
--- a/ckanext/xloader/jobs.py
+++ b/ckanext/xloader/jobs.py
@@ -247,7 +247,6 @@ def xloader_data_into_datastore_(input, job_dict):
     # Load it
     logger.info('Loading CSV')
     try:
-        import pdb; pdb.set_trace()
         fields = loader.load_csv(
             tmp_file.name,
             resource_id=resource['id'],
--- a/ckanext/xloader/tests/test_jobs.py
+++ b/ckanext/xloader/tests/test_jobs.py
@@ -262,52 +262,6 @@ class TestxloaderDataIntoDatastore(util.PluginsMixin):
 
     @mock_actions
     @responses.activate
-    def test_invalid_byte_sequence(self):
-        self.register_urls(filename='go-realtime.xlsx')
-        # This xlsx throws an Postgres error on INSERT because of
-        # 'invalid byte sequence for encoding "UTF8": 0x00' which causes
-        # the COPY to throw a psycopg2.DataError and umlauts in the file can
-        # cause problems for logging the error. We need to check that
-        # it correctly reverts to using messytables to load it
-        data = {
-            'api_key': self.api_key,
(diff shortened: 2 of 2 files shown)


stefina:
Thank you so much @davidread for following up on this! I haven't gotten around yet to provide a proper test so no worries that it was lost.
My fix just sugarcoats the failures that should be debugged in messytables but now the cases that worked before still work with xloader as the fallback is more likely to step in :)

Thank you!

stefina:
Thank you so much for following up on this @davidread !
I haven't got around to provide a proper test to show why our xlsx fails, so no worries that it got lost.
My fix just sugarcoats some errors which should be debugged in messytables I suppose. Until I have time to debug that this fix helps us a lot to make sure that the examples that worked before will still work.

Thanks a lot!

stefina:
Thank you so much for following up on this @davidread !
I haven't got around to provide a proper test to show why our xlsx fails, so no worries that it got lost.
My fix just sugarcoats some errors which should be debugged in messytables I suppose. Until I have time to debug that this fix helps us a lot to make sure that the examples that worked before will still work.

Thanks a lot!

stefina:
Thank you so much for following up on this @davidread !
I haven't got around to provide a proper test to show why our xlsx fails, so no worries that it got lost.
My fix just sugarcoats some errors which should be debugged in messytables I suppose. Until I have time to debug that this fix helps us a lot to make sure that the examples that worked before will still work.

Thanks a lot!

stefina:
Thank you so much for following up on this @davidread !
I haven't got around to provide a proper test to show why our xlsx fails, so no worries that it got lost.
My fix just sugarcoats some errors which should be debugged in messytables I suppose. Until I have time to debug that this fix helps us a lot to make sure that the examples that worked before will still work.

Thanks a lot!

PR #54, opened 2018-10-26
"Calculate record count" param added when using messytables

davidread:
This change is cherry-picked from https://github.com/ckan/datapusher/pull/174 - see there for more details.


ThrawnCA:
Does this have any relevance now that Messytables has been replaced with Tabulator?

davidread:
I have no interest in this any more, so will close

PR #57, opened 2018-11-07
56 keep processing the remaining resources when an error occurs

### LINKED ISSUE #56: Submit fails for dataset(s) when an upload for a single resource fails
When submitting all datasets to be xloadered and a resource can't be processed, the whole process is disrupted with an exception.

For context: We experienced this when a resource failed due to an encoding problem in its URL.
`paster xloader submit dataset -c /etc/ckan/default/production.ini`

We would still expect the command to keep processing the remaining resources (and furthermore all datasets, when using the submit_all-command).

stefina:
This is a suggestion to fix #56


davidread:
Thanks @stefina!

PR #59, opened 2018-11-07
refactor error-message to mention xloader instead of datapusher

stefina:
This is a tiny change that just changes the error-messages to be less confusing (because it was to me 😅 )


davidread:
Nice one :)

PR #63, opened 2018-11-14
Add option to load a minimum of rows

stefina:
This is my very rookie and WIP approach to make it possible to load only a part of a csv into the datastore if the file is bigger than the configured `max_content_length`

It will try to upload the first few configured number of rows if the file is too big.

It is lacking tests as this is so far a POC to discuss if this is beneficial and how this should be configured.


ckanext/xloader/jobs.py
```diff
@@ -190,26 +192,40 @@ def get_url():
...
+        tmp_file = get_tmp_file(url)
         length = 0
         m = hashlib.md5()
         for chunk in response.iter_content(CHUNK_SIZE):
             length += len(chunk)
             if length > MAX_CONTENT_LENGTH:
-                raise JobError(
-                    'Resource too large to process: {cl} > max ({max_cl}).'
-                    .format(cl=length, max_cl=MAX_CONTENT_LENGTH))
+                raise ResourceTooBigError
```
davidread: Can this be `DataTooBigError` to avoid confusion with CKAN's resources?

ckanext/xloader/jobs.py
```diff
@@ -281,6 +297,13 @@ def get_url():
     logger.info('Express Load completed')
 
 
+def get_tmp_file(url):
+    # download the file to a tempfile on disk
```
davidread: This comment should be kept with where it was called, rather than this function.

ckanext/xloader/jobs.py
davidread: Can this be `DataTooBigError` to avoid confusion with CKAN's resources?

ckanext/xloader/jobs.py
```diff
@@ -190,26 +192,40 @@ def get_url():
...
             tmp_file.write(chunk)
             m.update(chunk)
 
+    except ResourceTooBigError:
+        message = 'Resource too large to download: ' \
+                    '{cl} > max ({max_cl}).' \
+                    .format(cl=cl, max_cl=MAX_CONTENT_LENGTH)
+        if UPLOAD_EXCERPT:
+            logger.info(message)
+            logger.info('Uploading header and first {lines} lines to '
```
davidread: Can we use 'load' rather than 'upload', generally please?

ckanext/xloader/jobs.py
```diff
@@ -190,26 +192,40 @@ def get_url():
...
             if length > MAX_CONTENT_LENGTH:
-                raise JobError(
-                    'Resource too large to process: {cl} > max ({max_cl}).'
-                    .format(cl=length, max_cl=MAX_CONTENT_LENGTH))
+                raise ResourceTooBigError
             tmp_file.write(chunk)
             m.update(chunk)
 
+    except ResourceTooBigError:
+        message = 'Resource too large to download: ' \
```
davidread: My words, but maybe this is clearer as 'Data too large to load into Datastore'

stefina:
Thank you for your review and your input!
I refactored it quite a bit now, so that we use only one option to configure, if an excerpt should be loaded. 
The file will now be loaded line by line and if at any point during reading it, the `ckanext.xloader.max_content_length` is exceeded, a DataTooBigError will be thrown.  If `ckanext.xloader.upload_excerpt` is set to True Xloader will try to load the data, that is already written to the tmp_file (approximately data of about `ckanext.xloader.max_content_length`) to DataStore. 
To show the user, whether the datastore contains the full source or just an excerpt, `datastore_type` will be set to `full` or `excerpt` on the resource.

davidread:
Lots of good here.

However I still don't get why you want to load a large chunk, rather than say 100 rows. If it's for preview then you only really need the first page, but maybe I mis-guessed the use case. So, please can you say what the use case for this PR is (motivation)?

Also some terminology requests:
* 'upload' change to 'load' to fit in
* 'datastore_type' is too abstract. 'contains_all_the_records_of_the_source_file'=true/false is clearer. Happy for something shorter if it is as clear. BTW I didn't mean to suggest saving the datastore_active value, I meant to mean that you might set a new key-value by reusing the code that sets the datastore_active value. I've only looked briefly, but are you sure it's being saved to the db currently?

stefina:
Yeah, well, our use-case is indeed to still be able to show a preview even if the file-size exceeds the `max_content_length`. This way it seemed to be less complicated to understand and to configure to me. But I will re-consider this once I have a little more time on my hands next week :)
I do like this way though, as it enables us to load more into the DataStore if we have a lot of space, just not enough to load each resource as a whole.

Similar for `datastore_type` which seems like a good fit as the meaning fits conveniently to `resource_type`, will have to think about this, too.
The value `datastore_type` can be seen in the API for resource but not in the DB. But neither is `datastore_active`. Should they be saved in the DB? (Or maybe I just haven't found `datastore_active`)

I will definitely give all this more thought next week :) Thank you for your helpful input!

davidread:
Thanks for this.

Yeah, the name `resource_type` is something I don't like - `type` might mean so many things/dimensions. What is the "type" of say a CSV file containing of spending in Frankfurt in May 2015? There is not an obvious answer. But I might expect "CSV" or "spend" or "monthly". I probably wouldn't expect "incomplete". Much better to say what dimension you're talking about, in my opinion.

stefina:
Finally getting back to this.

I do understand your concern regarding the naming of `datastore_type`. Will try to come up of something more clear.

Regarding the "limit by a configured line-count" vs. "limit by content_length" (`ckanext.xloader.max_content_length`):
I am still drawn to the "limit by content_length"-option as one line could possibly contain a lot of data, too. In that case we would still have control in how much data we want to be loaded into the datastore.
Plus, as I said earlier, it enables us to load more into the DataStore if we have a lot of space, just not enough to load each resource as a whole.

What do you think?

davidread:
@stefina you make a good point about long rows. Happy to go with the "limit by content_length"-option then.

ckanext/xloader/jobs.py
```diff
@@ -159,57 +162,61 @@ def xloader_data_into_datastore_(input, job_dict):
...
-                    'Resource too large to process: {cl} > max ({max_cl}).'
-                    .format(cl=length, max_cl=MAX_CONTENT_LENGTH))
+                raise DataTooBigError
             tmp_file.write(chunk)
             m.update(chunk)
+        data['datastore_contains_all_records_of_source_file'] = True
 
+    except DataTooBigError:
+        tmp_file.close()
+        cl = response.headers.get('content-length')
```
metaodi: No need to do this again, `cl` is already set above.

ckanext/xloader/jobs.py
```diff
@@ -159,57 +162,61 @@ def xloader_data_into_datastore_(input, job_dict):
...
+        message = 'Data too large to load into Datastore: ' \
+                    '{cl} bytes > max {max_cl} bytes.' \
+                    .format(cl=cl or length, max_cl=MAX_CONTENT_LENGTH)
+        logger.warning(message)
+        file_format = resource.get('format')
```
metaodi: Is the `format` already set here? What are possible values? I guess it could either be user submitted or guessed by CKAN. Make sure to account for that (e.g. check for lowercase values, check for mimetypes).

ckanext/xloader/jobs.py
```diff
@@ -159,57 +162,61 @@ def xloader_data_into_datastore_(input, job_dict):
...
+        if MAX_EXCERPT_LINES > 0:
```
metaodi: Instead of indenting this whole block, I would add a guard condition:

```
if MAX_EXCERPT_LINES <= 0:
    raise JobError(message)
```

Like that, the rest of the code can be un-indented by 1 level and it makes the code better readable.

ckanext/xloader/jobs.py
```diff
@@ -159,57 +162,61 @@ def xloader_data_into_datastore_(input, job_dict):
...
+            for line in response.iter_lines(CHUNK_SIZE):
+                tmp_file.write(line + '\n')
+                m.update(line)
+                length += len(line)
+                line_count += 1
+                if length > MAX_CONTENT_LENGTH or line_count == MAX_EXCERPT_LINES:
+                    break
+            data['datastore_contains_all_records_of_source_file'] = False
+        else:
+            raise JobError(message)
```
metaodi: see above, get rid of this "else" by using a guard condition.

ckanext/xloader/jobs.py
```diff
@@ -285,16 +292,46 @@ def get_url():
     logger.info('Express Load completed')
 
 
+def get_response(url, headers):
+    def get_url():
```
metaodi: I would get rid of this extra method (`get_url`), since now you already created the `get_response` method.

ckanext/xloader/jobs.py
```diff
@@ -159,57 +162,61 @@ def xloader_data_into_datastore_(input, job_dict):
...
+            response = get_response(url, headers)
+            length = 0
+            line_count = 0
+            m = hashlib.md5()
+            for line in response.iter_lines(CHUNK_SIZE):
+                tmp_file.write(line + '\n')
+                m.update(line)
+                length += len(line)
+                line_count += 1
+                if length > MAX_CONTENT_LENGTH or line_count == MAX_EXCERPT_LINES:
```
metaodi: Not really an issue, but I would use `line_count >= MAX_EXCERPT_LINES` just to be safe. Right now it's not possible to get past this, but if for whatever reason you get two lines at once, this check would fail.

ckanext/xloader/jobs.py
```diff
@@ -324,7 +363,7 @@ def callback_xloader_hook(result_url, api_key, job_dict):
     return result.status_code == requests.codes.ok
 
 
-def set_datastore_active_flag(data_dict, flag):
+def set_resource_metadata(data_dict, flag):
```
davidread: I guess it's a bit inconsistent between how we pass datastore_active and datastore_contains_all_records_of_source_file now. How about we get rid of `flag` and just pass `update_dict` containing the two values?

ckanext/xloader/loader.py
```diff
@@ -56,6 +56,8 @@ def load_csv(csv_filepath, resource_id, mimetype='text/csv', logger=None):
             #                                          extension=format)
             # except Exception:
                 raise LoaderError('Messytables error: {}'.format(e))
+        except Exception:
+            raise FileCouldNotBeLoadedError()
```
davidread: I don't quite get this, I'm afraid - probably needs a comment to explain. If it raised LoaderError here then the details of the error would be logged for the user and it would try loading again with messytables, including 'trying the format' (Or maybe you've checked and it doesn't in this case?). Raising FileCouldNotBeLoadedError quits with: "Loading file raised an error: FileCouldNotBeLoadedError", doesn't it?

stefina: Yeah, this part is still not making me happy. The issue is, that at this point, we could come along with an excerpt of a a xls-file for example as we can't really rely on the `resource.get('format')` as it can be empty or even worse - contains wrong information altogether, which will then result in a corrupt file that can be processed. 
I wanted to prevent a second try in that case as the logging would mislead the user to what really went wrong (we would just end with the statement 'No entries found - nothing to load')
That is why I am now tempted to drop `EXCERPT_FORMATS` to begin with and deal with the error that will be thrown in the case of a corrupt file. Not sure if the error 'array index out of range' is a guarantee for the error that the excerpt produced though :) My next push hopefully makes this a little bit clearer.

ckanext/xloader/jobs.py
stefina: I agree, I changed it accordingly

ckanext/xloader/jobs.py
stefina: I removed the part altogether as it probably helps me almost never :) Using a more pythonic approach now and deal with the error later.

test.ini
```diff
@@ -17,6 +17,8 @@ solr_url = http://127.0.0.1:8983/solr
 # tests here.
 ckan.plugins = xloader
 ckanext.xloader.jobs_db.uri = sqlite:////tmp/jobs.db
+ckanext.xloader.max_content_length = 70000
+ckanext.xloader.max_excerpt_lines = 100
```
stefina: I hope there is a better way to mock these configuration-options in my tests, haven't gotten around to find that out. 
This is meant to demonstrate, though, how the tests and most of all the functionality works now.

stefina:
@davidread the tests are now ready with mocked globals.
Would appreciate a code review :)

Will remove the WIP-status, too, as this is getting closer to hopefully being able to be merged 🤞

davidread:
(Merge if you're happy.)

stefina:
Thank you for your review and helpful feedback you both @metaodi and @davidread 

Will go on and merge it then 🎉

PR #85, opened 2019-11-21
add paster command to add existing column types to the Data Dictionary

### LINKED ISSUE #84: Import existing column types into data dictionary?
We have a CKAN system that has been using the DataPusher, and we want to migrate to XLoader to improve performance and reduce administration overheads. However, we have thousands of resources parsed by messytables, and we want to preserve the type information (to format the data better for client applications), which XLoader doesn't by default.

To avoid updating each resource manually, we're looking at writing a paster command to inspect the existing columns and add Data Dictionary overrides for all of the non-text fields for when we move to XLoader.

Are there any obvious pitfalls with this approach?

ThrawnCA:
This resolves #84

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -98,7 +98,8 @@ Caveats
   resource's Data Dictionary tab (CKAN 2.7 onwards) to change these to numeric
   or datestamp and re-load the file. There is scope to do this automatically in
   future.
- * If you have existing data, you can pre-populate the Data Dictionary::
+
+  * If you have existing data, you can pre-populate the Data Dictionary:
 
      paster --plugin=ckanext-xloader migrate_types -c /etc/ckan/default/ckan.ini
 
--- a/README.rst
+++ b/README.rst
@@ -98,7 +98,7 @@ Caveats
   resource's Data Dictionary tab (CKAN 2.7 onwards) to change these to numeric
   or datestamp and re-load the file. There is scope to do this automatically in
   future.
-** If you have existing data, you can pre-populate the Data Dictionary:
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -99,6 +99,10 @@ Caveats
   or datestamp and re-load the file. There is scope to do this automatically in
   future.
 
+  * If you have existing data, you can pre-populate the Data Dictionary:
+
+     paster --plugin=ckanext-xloader migrate_types -c /etc/ckan/default/ckan.ini
```
davidread: Would you mind moving this to the Upgrading from DataPusher section, say before step 2?

ckanext/xloader/cli.py
```diff
@@ -230,3 +233,108 @@ def _print_status(self):
...
+            return
+
+        fields = []
+        for field in data_dict:
+            if field['type'] == 'text' and not self.options.include_text:
+                type_override = ''
+            else:
+                type_override = field['type']
+
+            if not 'info' in field:
```
davidread:
```suggestion
            if 'info' not in field:
```
and just some other minor flake8 errors to fix in this file, to make Travis pass

README.rst
ThrawnCA: Can do, probably on Monday.

ckanext/xloader/cli.py
ThrawnCA: This is done.

duttonw:
Looks good

davidread:
Thanks both!

Issue #90, opened 2019-12-03
Non-intuitive messytables behavior


ThrawnCA:
In testing XLoader and the Data Dictionary recently, I noticed that after setting just one column type to 'timestamp' on a new resource and re-submitting it, the whole Data Dictionary was populated with appropriate types. Checking the Datastore log, I realised that this was because our non-ISO dates (dd/mm/yyyy) had caused a direct PostgreSQL COPY to fail, so it fell back to using messytables - which meant that it guessed all of our types.

Although this behaviour is actually rather useful, it seems rather inconsistent that you get full type-guessing by *providing invalid input*, yet no such guessing if your data is entirely valid. It's also unfortunate that it presumably replaces any Data Dictionary overrides you may have entered manually.

So, there are two issues:

- If there are any existing Data Dictionary overrides, then messytables probably should not be permitted to overwrite them.

- How easily could the use of messytables be made configurable? It occurs to me that it would be useful to have at least a setting to use messytables on first creation of a resource, while attempting a fast COPY for further updates. I'm sure that there are also a lot of potential XLoader users (possibly even my own workplace) who would prefer the maximum-backwards-compatibility option of *always* using messytables, despite the more modest speed gains compared to DataPusher - and as a side benefit, this wouldn't result in the log warning when PostgreSQL COPY fails (which can confuse non-technical users).

ThrawnCA:
Looks like #36 already addressed the first point.

davidread:
Good points. This is all an excellent area to improve, it's not something me or my client are actively pursuing, so I'm happy for you to do this. 

Definitely when messytables loads it should respect the Data Dictionary. 

On detecting types in general use, I'm not keen to rely too much on the messytables load because it does a tiresome conversion to chunks of JSON and then INSERT statements - requires a lot of CPU and it runs on the same machine as CKAN, which was not necessarily the case with DataPusher's separate queue.

However we might still use messytables' type detection ability. I think it scans the first chunk of the CSV and makes a simple decision - I can't remember what the basis is. We could try setting the column types on that basis (the `datastore_create`) and try the COPY, and if that fails revert to string columns and redo the COPY?

A better alternative might be to still load initially as strings, then use messytables or some other heuristic to guess types, and then see if postgres itself can successfully cast a column to that type. If it can, we set the Data Dictionary for next time. And if next time it fails, then maybe it reverts to strings, COPY and then try the casting thing again.

[...]

PR #92, opened 2019-12-04
add compatibility mode flag, #90

ThrawnCA:
Setting `ckanext.xloader.compatibility_mode` to True will result in automatically using messytables, instead of attempting a PostgreSQL COPY first.

This caters for clients who prefer simpler processes and a more seamless migration from DataPusher instead of raw upload speed, or who know that their data will not be handled by COPY (eg non-ISO dates).

If the property is not set, it defaults to False ie the previous behavior, attempting COPY first.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,13 @@
+0.5.0 2019-12-04
+================
+
+Features:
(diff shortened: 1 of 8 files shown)


davidread:
Thanks for this. I just did a bit of [renaming](https://github.com/ckan/ckanext-xloader/pull/96) - hope you don't mind.

PR #100, opened 2020-02-22
Update travis config from the ckan template

davidread:


ThrawnCA:
Travis CI is no longer in use.

PR #111, opened 2020-08-17
add support for unicode header names

OriHoch:
Currently, unicode characters in headers are encoded to ascii using unidecode library.

This PR adds a config option (`ckanext.xloader.unicode_headers`) which allows to keep unicode charactes in the column names

Includes new tests for this functionality + verified no regressions in existing tests


ckanext/xloader/loader.py
```diff
@@ -37,7 +37,7 @@ def get_write_engine():
 MAX_COLUMN_LENGTH = 63
 
 
-def load_csv(csv_filepath, resource_id, mimetype='text/csv', logger=None):
+def load_csv(csv_filepath, resource_id, mimetype='text/csv', logger=None, unicode_headers=None):
```
davidread, line 40: It looks like `unicode_headers` is a parameter only used for test purposes, which is something I'd really prefer to avoid. How about the test instead uses `helpers.change_config` to achieve the same thing, and drop this param? https://docs.ckan.org/en/latest/extensions/testing-extensions.html

OriHoch: yes, that's exactly what I was looking for, I'll change it

kaplanlior:
@davidread @OriHoch 

First, thanks for the PR and the review.
Can we move forward with merging it ?

Anything needed except a rebase of the branch with HEAD?

ThrawnCA:
This looks good to me, but @OriHoch can you resolve the merge conflict?

duttonw:
@ThrawnCA are you able to cherry pick his work and make it function. (the file that they wanted to put under test and see the result on load)

PR #121, opened 2020-11-05
Update 'submit all' CLI command to include private datasets

michael-long88:
Updated the CLI command to submit all packages to the xloader to include both private and public datasets.


michael-long88:
Could someone take a look at this? It looks like the tests are failing, but not from anything on our end.

PR #136, opened 2021-07-16
Add ssl_verify option to callback_xloader_hook

pdelboca:
When working locally and using self signed certificates the POST request to `xloader_hook` fails to verify the certificate. This makes the job to never be updated to complete status after finishing.

I'm adding the `verify=SSL_VERIFY` parameter as we have in the `requests.get(...)` method.


kowh-ai (approved):
Nice @pdelboca - I have merged this PR

PR #140, opened 2021-09-01
fix xloader exceptions

TomeCirun:
Enabling xloader causes an exception at the end of the dataset creation process, when submitting the resources.

image

So .. i was following @amercader 's  work on [5137](https://github.com/ckan/ckan/pull/5173)


tino097 (approved):
LGTM

PR #141, opened 2021-09-22
Fix xloader status timestamps

amercader:
Encode timestamps as strings on `xloader_status` call otherwise there is an error when trying to return the API JSON value.

This also includes a small fix to correctly report exceptions in stdout (40da6c6d8224907ad4798e0b4f97cdea64f65ce5)


tino097 (approved):
LGTM

PR #144, opened 2021-11-08
support replacable CKAN hostname

aruneko:
When we run CKAN and XLoader on a Docker environment, we run only one process in a docker container as usual. Thus, we make two containers for that, and these containers have their IP addresses.

When XLoader calls the API endpoint `/api/3/action/resource_show` on CKAN, `url` value in the returned JSON object starts from `http://localhost` based on the congifuration `ckan.site_url`.

But `localhost` of the XLoader container doesn't point to the CKAN container, since these processes are running on another container. The hostname should be equal to the CKAN container name by default.

Therefore, the hostname of the `url` value should be replaceable via the configuration file.

That's why I added a new configuration `ckanext.xloader.ckan_host` to resolve the problem. This change should not affect the existing environment.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -265,6 +265,11 @@ Configuration:
     # to True.
     ckanext.xloader.ssl_verify = True
 
+    # The hostname of a CKAN server. The value should have the hostname.
+    # And you can add the port number on demand.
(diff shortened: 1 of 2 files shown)


aruneko:
hmm, I don't think about the link functionality. At once I close this pull request.

PR #150, opened 2021-12-06
add future to requirements.txt

RabiaSajjad:
Python 3 support requires using `builtins` in commit https://github.com/open-data/ckanext-xloader/commit/8c43ceed879af9934dd843a36ab7c68d9a0535c8#diff-b0f564951458e386ad6106e20fa8938150638c260b1921aceba34dae89d04d58. Adding `future` to the requirements to avoid ImportError.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,17 +61,14 @@ jobs:
     - name: Setup extension (CKAN >= 2.9)
       if: ${{ matrix.ckan-version != '2.7' && matrix.ckan-version != '2.8' }}
       run: |
-        pip install future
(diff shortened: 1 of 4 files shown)


smotornyuk:
While using `future` in our projects I've run into a couple of conflicts caused by the following structure: 
```
from future import standard_library
standard_library.install_aliases()
```

It does a lot of monkey-patching and there are py3 libraries, that cannot accept this:) I was the one, who initially added `feature` dependency and I must admit, that it was a bad idea. Can you try to get rid of `feature` instead? using `six.text_type` and wrapping `zip`/`map`/etc into `list` must be sufficient, i believe

PR #155, opened 2022-03-03
Add BS5 support

TomeCirun:
Bootstrap 5 has changed the naming of the data attributes that are generally used by its components that use JavaScript as part of their functionality.

Previously, most components that relied on some JavaScript functionalities would have data attributes starting with data-.
In Bootstrap 5, data attributes for these components now start with data-bs to easily namespace Bootstrap attributes.


tino097 (approved):

ckanext/xloader/templates/xloader/resource_data.html
```diff
@@ -73,7 +73,7 @@ <h3>{{ _('Upload Log') }}</h3>
             {% endfor %}
             <span class="date" title="{{ h.render_datetime(item.timestamp, with_hours=True) }}">
               {{ h.time_ago_from_timestamp(item.timestamp) }}
-              <a href="#" data-target="popover" data-content="<dl>{% for key, value in item.items() %}<dt>{{ key }}</dt><dd>{{ h.clean_html(value|string) }}</dd>{% endfor %}</dl>" data-html="true">{{ _('Details') }}</a>
+              <a href="#" data-bs-target="popover" data-bs-content="<dl>{% for key, value in item.items() %}<dt>{{ key }}</dt><dd>{{ h.clean_html(value|string) }}</dd>{% endfor %}</dl>" data-bs-html="true">{{ _('Details') }}</a>
```
amercader, line 76: Similarly to https://github.com/ckan/ckanext-harvest/pull/492, let's just add both versions of the data params rather than duplicate all templates

pdelboca:
Thanks @TomeCirun ! 

I'm merging this so I can keep working on #156

PR #156, opened 2022-04-08
Add support for 2.10

pdelboca:
This PR is to support CKAN 2.10

Upstream ticket: https://github.com/ckan/ckan/issues/6787

## Main Changes
- Fixed SQLAlchemy syntax that fails in `1.4`
- Add backwards compatibility for IResourceController methods.
- Add `conftest.py` file so we can use custom fixtures.

### DIFF
--- a/ckanext/xloader/plugin.py
+++ b/ckanext/xloader/plugin.py
@@ -69,13 +69,6 @@ class xloaderPlugin(plugins.SingletonPlugin):
                 action='resource_data', ckan_icon='cloud-upload')
             return m
 
-    # IResourceController
-
-    def before_show(self, resource_dict):
-        resource_dict[
-            'datastore_contains_all_records_of_source_file'] = toolkit.asbool(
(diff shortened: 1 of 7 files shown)


ckanext/xloader/plugin.py
```diff
@@ -123,11 +116,23 @@ def notify(self, resource):
         )
         self._submit_to_xloader(resource_dict)
 
-    # IResourceController                
+    # IResourceController
+    if toolkit.check_ckan_version('2.10'):
```
amercader, line 120: In this case the logic in the hooks is very simple so it it ok, but in general I think this is a better pattern to avoid duplication:

```python
class MyPlugin():

    # CKAN < 2.10
    def after_create(self, context, data_dict):
        return self.after_resource_create(context, data_dict)

    # CKAN >= 2.10
    def after_resource_create(self, context, data_dict):
        # Do whatever

```

PR #157, opened 2022-04-22
Raise loadererror when the delimiter cannot be detected

ba-tno:
In case messytables by default is off and one supplies a non-CSV file, the delimiter detection fails with a UnicodeDecodeError as it tries to parse the binary data (I presume). By catching that error and raising a LoadError instead, subsequent code picks it up and uses messytables as a fallback option.


pdelboca:
Thanks @ba-tno !

PR #168, opened 2022-08-17
Limit duplicate builds

ThrawnCA:
Don't force an extra GitHub Actions workflow execution on pull request unless the target branch is 'master'.

This will reduce the number of duplicate executions (since we already trigger on 'push'), while still ensuring that cross-repo pull requests to 'master' will be tested.


pdelboca:
Thanks @ThrawnCA !

PR #180, opened 2023-02-22
Drop support for old CKAN versions and add CSRF support

pdelboca:
This PR deprecates support for 2.7 and 2.8 to match our policy: https://docs.ckan.org/en/latest/maintaining/releases.html#supported-versions

In addtion
 - Adds support for CSRF token
 - Adds `config_declaration.yaml` file


setup.py
```diff
@@ -15,7 +15,7 @@
     # Versions should comply with PEP440.  For a discussion on single-sourcing
     # the version across setup.py and the project code, see
     # http://packaging.python.org/en/latest/tutorial.html#version
-    version='0.12.2',
+    version='0.13.0',
```
ThrawnCA, line 18: Since this is dropping support for previous APIs, it should probably be a major version update, not minor.

ckanext/xloader/plugin.py
```diff
@@ -8,6 +8,12 @@
 from . import action, auth, helpers as xloader_helpers, utils
 from .loader import fulltext_function_exists, get_write_engine
 
+IConfigDeclaration = None
+if toolkit.check_ckan_version("2.10"):
+    import yaml
+    import os
+    IConfigDeclaration = plugins.IConfigDeclaration
+
```
smotornyuk, line 16: Sorry, I mentioned it on the call but it took time before I finally managed to add a comment here.

You can simplify adding declarations using blanket:
```
try:
    config_declarations = tk.blanket.config_declarations
except AttributeError:
    config_declarations = lambda cls: cls

@config_declarations
class xloaderPlugin(plugins.SingletonPlugin):
...
```

As you see, here we define a no-op decorator if the blanket is unavailable so that it won't produce an error in 2.9. In this way you can completely remove the implementation of the interface from the class because it looks for `config_declaration.yaml` inside `ckanext/<plugin>` by default

ckanext/xloader/config_declaration.yaml
```diff
@@ -0,0 +1,116 @@
+version: 1
+groups:
+  - annotation: ckanext-xloader settings
+    options:
+      - key: ckanext.xloader.jobs_db.uri
+        example: sqlite:////tmp/xloader_jobs.db
```
smotornyuk, line 6: At the moment [it's a default value](https://github.com/ckan/ckanext-xloader/blob/9db532439dbb8b88f7c9d11f1e573ff6c164e3b9/ckanext/xloader/db.py#L40). Can we leave this default, as it shouldn't do any harm.

ckanext/xloader/config_declaration.yaml
```diff
@@ -0,0 +1,116 @@
...
+      - key: ckanext.xloader.formats
+        example: csv application/csv xls application/vnd.ms-excel
+        description: |
+            The formats that are accepted. If the value of the resource.format is
+            anything else then it won't be 'xloadered' to DataStore (and will therefore
+            only be available to users in the form of the original download/link).
+            Case insensitive. Defaults are listed in plugin.py.
+        required: false
+      - key: ckanext.xloader.max_content_length
+        default: 1000000000
```
smotornyuk, line 29: I'd say that using `_` simplifies reading such values. `default: 1_000_000_000` is a valid value for the `int` type.

setup.py
pdelboca: Good catch @ThrawnCA ! I'm fixing this and proposing a `1.0.0` version.

PR #181, opened 2023-03-02
Refactor test_jobs.py

pdelboca:
The current implementation is skipping `test_jobs.py` given its old and deprecated engineering.

This PR gives some love by refactoring them using modern CKAN fixtures.

@smotornyuk this is a small follow up of #180  that I have been working on this morning.

### DIFF
--- a/ckanext/xloader/tests/test_jobs.py
+++ b/ckanext/xloader/tests/test_jobs.py
@@ -12,6 +12,7 @@ from ckan.tests import helpers, factories
 from unittest import mock
 
 from ckanext.xloader import jobs
+from ckanext.xloader.utils import get_xloader_user_apitoken
 
 
 _TEST_FILE_CONTENT = "x, y\n1,2\n2,4\n3,6\n4,8\n5,10"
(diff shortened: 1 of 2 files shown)


ckanext/xloader/tests/test_jobs.py
```diff
@@ -1,740 +1,131 @@
...
+    resp.headers = {'content-length': 2000000000}
+    return resp
+
+
+@pytest.fixture
+def apikey():
+    try:
+        sysadmin = factories.SysadminWithToken()
+    except AttributeError:
+        # To provide support with CKAN 2.9
```
ThrawnCA, line 42: Wouldn't `hasattr` be better, to avoid throwing and catching exceptions? Or even toolkit `check_ckan_version`

ThrawnCA (changes requested):

ckanext/xloader/tests/test_jobs.py
smotornyuk: Don't start [this holy war](https://docs.python.org/3.6/glossary.html#term-eafp) 🤣 

It's not a performance-critical fixture, so let's spare Patricio and leave the initial version here.

pdelboca: I got curious: https://gist.github.com/pdelboca/bc5dc12a10718da5559be360437e280b

smotornyuk: Yeah, but it would be nice to mention that in your case numbers are approximately the same because of the complexity of `check_ckan_version`. I believe, for general cases when you are checking something plain, like `if 1 < 2:`, `if` is considerably faster. 
It doesn't change our situation, but it would be shame if someone finds your gist and use it to prove that `if` and `try` are the same all the time :)

ThrawnCA: Eh, I hadn't realised that EAFP is an officially approved style. I don't like it myself, but okay.

pdelboca: There isn't an approved style guide. I was just playing around because I got genuinely curious about it since there wasn't a clear definition of "better".

I refactored it to use `check_ckan_version`.

ThrawnCA (approved):
`check_ckan_version` is good for making the code self-documenting, at least.

Issue #182, opened 2023-03-30
Type guessing in 0.12 is less robust


ThrawnCA:
The new type guessing system in 0.12+ encounters errors that the old one didn't. For example, 
[QOLSVC-1280_current-program-funding-dtis_1_2_2_2_1.csv](https://github.com/ckan/ckanext-xloader/files/11107613/QOLSVC-1280_current-program-funding-dtis_1_2_2_2_1.csv) results in an error due to cells not being recognised as any type, even string:

    Error: File "/mnt/local_data/ckan_venv/src/ckanext-xloader/ckanext/xloader/utils.py", line 221, in type_guess _columns.append(max(guesses_tuples, key=lambda t_n: t_n[1])[0]) ValueError('max() arg is an empty sequence') 

Or when a field is recognised as numeric or timestamp, and then a later row leaves that field blank, as per 
[QOLSVC-1280_current-program-funding-dtis_1_2_2_2_2.csv](https://github.com/ckan/ckanext-xloader/files/11107621/QOLSVC-1280_current-program-funding-dtis_1_2_2_2_2.csv):

    Error: Validation error writing rows to db: None - {'message': 'The data was invalid: invalid input syntax for type timestamp: ""'} 

The old messytables-based type guessing had no trouble parsing these.

ThrawnCA:
It appears that the first problem arises when data is numeric on some rows but free text on other rows. The numeric rows cause `str` to be removed from the set of guessed types, and the string rows cause `Decimal` to be removed, leaving no guessed types at all.

ThrawnCA:
The "invalid input syntax for type timestamp" error appears to occur because the type guessing accepts an empty string as a valid timestamp value, but the SQL library does not.

ThrawnCA:
Found another issue, where specific data combinations will result in the parser ignoring double quotes and just splitting rows on every comma. I'm not sure exactly of the exact root cause yet, continuing to investigate.
[blp_dashboard_dataset_31-mar-2023.csv](https://github.com/ckan/ckanext-xloader/files/11571669/blp_dashboard_dataset_31-mar-2023.csv)
The attached file contains 2 rows and will trigger the error, but if you upload a file with just one of the rows - either one, it doesn't matter - then it succeeds.

ThrawnCA:
Ah, found the cause. The input file uses double quotes, but it encloses so many single quotes that the `csv` module assumes single quotes must be the quote character.

[...]

PR #185, opened 2023-04-17
[QOLDEV-347] apply 'str' fallback type correctly, #182

ThrawnCA:
- If all types have been rejected, ensure that the fallback flag is correctly set
- Ensure that empty strings are converted to None before attempting to turn them into SQL timestamps or numbers.


duttonw (approved):

ThrawnCA:
Will reopen on a dedicated branch.

PR #190, opened 2023-06-13
Conditional DataStore Tab in Resource Edit

JVickery-TBS:
feat(templates): conditional DataStore tab;

- Added helper for valid xloader formats.
- Added condition to display the DataStore tab in Resource Edit.

### DIFF
--- a/ckanext/xloader/helpers.py
+++ b/ckanext/xloader/helpers.py
@@ -25,3 +25,9 @@ def xloader_status_description(status):
         return captions.get(status['status'], status['status'].capitalize())
     else:
         return _('Not Uploaded Yet')
+
+
+def is_xloader_format(resource_format):
+    from ckanext.xloader.plugin import XLoaderFormats
+
+    return XLoaderFormats.is_it_an_xloader_format(resource_format)
--- a/ckanext/xloader/helpers.py
+++ b/ckanext/xloader/helpers.py
@@ -31,10 +31,8 @@ def xloader_status_description(status):
 def is_resource_supported_by_xloader(res_dict, check_access = True):
     is_supported_format = XLoaderFormats.is_it_an_xloader_format(res_dict.get('format'))
     is_datastore_active = res_dict.get('datastore_active', False)
-    if check_access:
-        user_has_access = toolkit.h.check_access('package_update', {'id':res_dict.get('package_id')})
-    else:
-        user_has_access = True
+    user_has_access = not check_access or toolkit.h.check_access('package_update',
+                                                                 {'id':res_dict.get('package_id')})
     try:
         is_supported_url_type = res_dict.get('url_type') not in toolkit.h.datastore_rw_resource_url_types()
     except AttributeError:
--- a/ckanext/xloader/helpers.py
+++ b/ckanext/xloader/helpers.py
@@ -1,4 +1,5 @@
 import ckan.plugins.toolkit as toolkit
+from ckanext.xloader.utils import XLoaderFormats
 
 
 def xloader_status(resource_id):
@@ -27,14 +28,15 @@ def xloader_status_description(status):
         return _('Not Uploaded Yet')
 
 
-def is_xloader_format(resource_format):
-    from ckanext.xloader.plugin import XLoaderFormats
-
-    return XLoaderFormats.is_it_an_xloader_format(resource_format)
-
-
-def is_xloader_type(resource_url_type):
+def is_resource_supported_by_xloader(res_dict, check_access = True):
+    is_supported_format = XLoaderFormats.is_it_an_xloader_format(res_dict.get('format'))
+    is_datastore_active = res_dict.get('datastore_active', False)
+    if check_access:
+        user_has_access = toolkit.h.check_access('package_update', {'id':res_dict.get('package_id')})
+    else:
+        user_has_access = True
     try:
-        return resource_url_type not in toolkit.h.datastore_rw_resource_url_types()
+        is_supported_url_type = res_dict.get('url_type') not in toolkit.h.datastore_rw_resource_url_types()
     except AttributeError:
-        return (resource_url_type == 'upload' or resource_url_type == '')
+        is_supported_url_type = (res_dict.get('url_type') == 'upload' or res_dict.get('url_type') == '')
+    return (is_supported_format or is_datastore_active) and user_has_access and is_supported_url_type
--- a/ckanext/xloader/helpers.py
+++ b/ckanext/xloader/helpers.py
@@ -31,3 +31,10 @@ def is_xloader_format(resource_format):
     from ckanext.xloader.plugin import XLoaderFormats
 
     return XLoaderFormats.is_it_an_xloader_format(resource_format)
+
+
+def is_xloader_type(resource_url_type):
+    try:
+        return resource_url_type not in toolkit.h.datastore_rw_resource_url_types()
+    except AttributeError:
+        return (resource_url_type == 'upload' or resource_url_type == '')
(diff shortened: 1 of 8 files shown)


ThrawnCA:
Looks good to me, but I'm not a maintainer here. Do you want to open a pull request to the qld-gov-au fork, ideally the develop branch?

Actually, while you're working on this, do you think you could add a datastore tab link to the resource view page when appropriately authorised? Would be nice to be able to go there in one click.

ThrawnCA (approved):

ckanext/xloader/templates/package/resource_read.html
```diff
@@ -0,0 +1,12 @@
+{% ckan_extends %}
+
+{% block action_manage %}
+  {{ super() }}
+  {%
+    if (h.is_xloader_format(res.format) or res.datastore_active)
+    and h.check_access('package_update', {'id':pkg.id })
+    and res.url_type not in h.datastore_rw_resource_url_types()
```
ThrawnCA, line 8: This helper does not exist in any currently released version of CKAN. We need to gracefully handle its absence.

ThrawnCA (changes requested):

ckanext/xloader/templates/package/resource_read.html
JVickery-TBS: last condition now replaced with new helper which calls `datastore_rw_resource_url_types` in try/catch.

JVickery-TBS:
DataStore action link now added to more templates as to feedback here: https://github.com/ckan/ckan/pull/7647#pullrequestreview-1480247877

ThrawnCA (changes requested):
See comments on https://github.com/qld-gov-au/ckanext-xloader/pull/49

ThrawnCA:
Don't get me wrong, this is a promising feature, I just want to help ensure it's at its best.

JVickery-TBS:
Comments addressed from https://github.com/qld-gov-au/ckanext-xloader/pull/49

ThrawnCA (approved):

ckanext/xloader/templates/package/resource_read.html
```diff
@@ -0,0 +1,8 @@
+{% ckan_extends %}
+
+{% block action_manage_inner %}
```
ThrawnCA, line 3: This is another "only in master" feature. On any released version, it will not take effect as this block does not exist in the parent template.

ckanext/xloader/templates/package/snippets/resource_item.html
```diff
@@ -0,0 +1,8 @@
+{% ckan_extends %}
+
+{% block resource_item_explore_inner %}
```
ThrawnCA, line 3: This block exists only in CKAN core master, not in any released version, and so it will have no effect.

ckanext/xloader/templates/package/snippets/resources.html
```diff
@@ -0,0 +1,8 @@
+{% ckan_extends %}
+
+{% block resources_list_edit_dropdown_inner %}
```
ThrawnCA, line 3: This block exists only in CKAN core master, not in any released version, and so it will have no effect.

ThrawnCA (changes requested):

JVickery-TBS:
@ThrawnCA this PR could just get marked for a release alongside the next CKAN release.

As the templates are made for the latest ckan, users can just override them if they need to for older releases.

ThrawnCA:
> @ThrawnCA this PR could just get marked for a release alongside the next CKAN release.

You could, but then no one can do any testing of it now. I intend to deploy this on data.qld.gov.au, but we're still on 2.9, currently working on the 2.10 upgrade. If even 2.10 can't use it... it'll be potentially a long time before we can give any feedback.

wardi:
In extensions it's desirable to be able to support a number of ckan releases. The template blocks won't be visible in current releases but do they prevent people from using the feature?

JVickery-TBS:
These pieces of templates were not in Xloader at all. These templates and blocks are completely new in this PR. So it does not affect the functionality of the extension at all.

The only template the existed before was `package/resource_edit_base.html` and the only change to that in this PR is the new conditional tab.

So all of the newly added templates could be considered "for the future"

ThrawnCA:
Why not just add the new link to the `resource_actions_inner` block instead? That one exists in both old and new CKAN versions.

ckanext/xloader/helpers.py
```diff
@@ -25,3 +26,17 @@ def xloader_status_description(status):
...
+    is_supported_format = XLoaderFormats.is_it_an_xloader_format(res_dict.get('format'))
+    is_datastore_active = res_dict.get('datastore_active', False)
+    if check_access:
+        user_has_access = toolkit.h.check_access('package_update', {'id':res_dict.get('package_id')})
+    else:
+        user_has_access = True
+    try:
+        is_supported_url_type = res_dict.get('url_type') not in toolkit.h.datastore_rw_resource_url_types()
+    except AttributeError:
+        is_supported_url_type = (res_dict.get('url_type') == 'upload' or res_dict.get('url_type') == '')
```
ThrawnCA, line 41: This should handle any falsy url_type the same way as empty string. It is possible for url_type to be `None`.

ThrawnCA (changes requested):

ckanext/xloader/templates/package/snippets/resources.html
JVickery-TBS: yeah there is no block in less than 2.10 that does anything in this template

ckanext/xloader/templates/package/snippets/resource_item.html
JVickery-TBS: Added max 2.10 version check in other block

ckanext/xloader/templates/package/resource_read.html
JVickery-TBS: added max 2.10 version check in other block now.

ckanext/xloader/helpers.py
JVickery-TBS: changed to falsy check instead of just blank string.

ckanext/xloader/templates/package/snippets/resources.html
ThrawnCA: Ah, my bad, I missed the fact that this block appears in the Datastore templates rather than core ones. So it's 2.10 only but otherwise okay.

ckanext/xloader/utils.py
```diff
@@ -9,6 +9,36 @@
...
+
+
+class XLoaderFormats(object):
+    formats = None
+
+    @classmethod
+    def is_it_an_xloader_format(cls, format_):
+        if cls.formats is None:
+            cls._formats = config.get("ckanext.xloader.formats")
+            if cls._formats is not None:
```
ThrawnCA, line 35: I assume we don't just use `if cls._formats:` because if the config is an empty list, we want to preserve that. Should there be a comment to explain it?

JVickery-TBS: @ThrawnCA yeah I believe so, I have tested with an empty list and this seems to work.

I did not write this class, just moved it from helpers to utils a while ago. But added in the in-line comment now

ckanext/xloader/helpers.py
```diff
@@ -25,3 +26,17 @@ def xloader_status_description(status):
         return captions.get(status['status'], status['status'].capitalize())
     else:
         return _('Not Uploaded Yet')
+
+
+def is_resource_supported_by_xloader(res_dict, check_access = True):
+    is_supported_format = XLoaderFormats.is_it_an_xloader_format(res_dict.get('format'))
+    is_datastore_active = res_dict.get('datastore_active', False)
+    if check_access:
```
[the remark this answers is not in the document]
JVickery-TBS: @ThrawnCA used short-hand now

ThrawnCA (approved):

duttonw (approved):

PR #193, opened 2023-08-17
Fix resource format KeyError

alexmorev:


ThrawnCA (approved):

ckanext/xloader/plugin.py
```diff
@@ -155,11 +155,12 @@ def after_update(self, context, resource_dict):
...
-        if not XLoaderFormats.is_it_an_xloader_format(resource_dict["format"]):
+        resource_format = resource_dict.get("format")
+        if not XLoaderFormats.is_it_an_xloader_format(resource_format):
             log.debug(
-                "Skipping xloading resource {id} because "
-                'format "{format}" is not configured to be '
-                "xloadered".format(**resource_dict)
+                "Skipping xloading resource {} because "
+                'format "{}" is not configured to be '
+                "xloadered".format(resource_dict["id"], resource_format)
```
ThrawnCA, line 163: @alexmorev This could be improved by using parameterised logging instead of `format`, and triple quotes so that we don't need to mix single and double quotes.

PR #195, opened 2023-11-01
Quality of life improvements

duttonw:
* Add Dependabot to provide updated library's and kick of testing on our behalf
* Make flake8 ignore less rules to improve coding standards
* Increase timeouts for non-imported records to handle super large datasets (1million + rows)
* Re add backward compatibility: If ckanext.xloader.use_type_guessing is not configured, fall back to
   deprecated config option ckanext.xloader.just_load_with_messytables 
* Include chardet 5.2.0 UniversalDetector to better sniff the file encoding to stop failed direct loads where confidence is over 70%
  Fallback to ISO_8859  which covers majority of Windows-1252
* Better handle empty timestamp and numeric columns, direct load fails to import if its an empty string vs required None object.
* Direct Import Load did not handle hh:mm correctly and imported them as time delta of current import time. 
   So we try to convert cells to numbers or timestamps if applicable:
       If a list of types was supplied, use that.  
       If not, then try converting each column to numeric first, then to a timestamp. 
       If both fail, just keep it as a string.
*  If resource is_resource_supported_by_xloader. Add nav button to datastore so you don't need to visit the 'manage/edit' to allow 
   when auth allows (Resource uploader and above)
* Performance Improvement: On DataStore Tab of resource, Truncate very long logs:
  Only show the start and end (default first 50 and last 50) with ability to request more rows or all rows if so desired. 

As well as lots of extra tests for edge conditions


duttonw:
Hi @kowh-ai , @smotornyuk 
Just wondering what your capacity is for looking of this improvement PR. 

Regards,

@duttonw

ThrawnCA (approved):

ThrawnCA (approved):

PR #196, opened 2023-11-08
Clean Datastore Tables Job

JVickery-TBS:
feat(jobs): added job to delete datastore tables;

- Added config option `clean_datastore_tables`.
- Added job to delete datastore tables from unsupported formats.

This adds the possible feature (controlled by config option) to delete datastore tables for unsupported formats. This is useful in the case that a user updates a Resource to no longer be a datastore resource, but the tables remain. So this helps keep the size of the datastore database down a bit


ckanext/xloader/plugin.py
```diff
@@ -4,6 +4,10 @@
 
 from ckan import plugins
 from ckan.plugins import toolkit
+from ckan.model.domain_object import DomainObjectOperation
+from ckan.model.resource import Resource
+
+import ckanapi
```
ThrawnCA, line 10: `toolkit` should already be able to do everything we need.

ckanext/xloader/plugin.py
```diff
@@ -212,3 +229,40 @@ def get_helpers(self):
...
         }
+
+
+def _should_remove_unsupported_resource_from_datastore(res_dict_or_obj):
+    if not toolkit.asbool(toolkit.config.get('ckanext.xloader.clean_datastore_tables', False)):
+        return False
+    if isinstance(res_dict_or_obj, Resource):
+        res_dict_or_obj = res_dict_or_obj.__dict__
+    return ((not XLoaderFormats.is_it_an_xloader_format(res_dict_or_obj.get('format', u''))
+                or res_dict_or_obj.get('url_changed', False))
```
ThrawnCA, line 240: What is the relevance of `url_changed` here?

ThrawnCA (changes requested):

ckanext/xloader/plugin.py
```diff
@@ -212,3 +229,40 @@ def get_helpers(self):
...
+    """
+    lc = ckanapi.LocalCKAN()
+    try:
+        res = lc.action.resource_show(id=resource_id)
+    except toolkit.ObjectNotFound:
+        log.error('Resource %s does not exist.' % res['id'])
+        return
+
+    if _should_remove_unsupported_resource_from_datastore(res):
+        log.info('Unsupported resource format "{}". Deleting datastore tables for resource {}'
```
ThrawnCA, line 262: This should use parameterised logging with `%s` instead of `format`.

JVickery-TBS: Done

ckanext/xloader/plugin.py
JVickery-TBS: @ThrawnCA same thing as other PR, just a specific thing in our setup to my new understanding. As generally, systems can fetch from external source. E.g. a CSV file hosted somewhere else can still be fetched by Xloader and loaded into the datastore?

ckanext/xloader/plugin.py
JVickery-TBS: Using `get_action` from toolkit and removed `ckanapi` dependency

ckanext/xloader/plugin.py
```diff
@@ -231,17 +230,14 @@ def get_helpers(self):
...
-        res_dict_or_obj = res_dict_or_obj.__dict__
-    return ((not XLoaderFormats.is_it_an_xloader_format(res_dict_or_obj.get('format', u''))
-                or res_dict_or_obj.get('url_changed', False))
-            and (res_dict_or_obj.get('url_type') == 'upload'
-                or res_dict_or_obj.get('url_type') == '')
-            and (res_dict_or_obj.get('datastore_active', False)
-                or res_dict_or_obj.get('extras', {}).get('datastore_active', False)))
+    return (not XLoaderFormats.is_it_an_xloader_format(res_dict.get('format', u''))
+            and (res_dict.get('url_type') == 'upload'
+                or res_dict.get('url_type') == '')
```
ThrawnCA, line 238: This should allow any falsy value. We've been tripped up by that before.

ckanext/xloader/plugin.py
```diff
@@ -251,18 +247,20 @@ def _remove_unsupported_resource_from_datastore(resource_id):
...
+        res = toolkit.get_action('resource_show')(context, {"id": resource_id})
     except toolkit.ObjectNotFound:
         log.error('Resource %s does not exist.' % res['id'])
         return
 
     if _should_remove_unsupported_resource_from_datastore(res):
-        log.info('Unsupported resource format "{}". Deleting datastore tables for resource {}'
-            .format(res.get(u'format', u'').lower(), res['id']))
+        log.info('Unsupported resource format "%s". Deleting datastore tables for resource %s'
+            % (res.get(u'format', u'').lower(), res['id']))
```
ThrawnCA, line 259: Don't use % based interpolation, just pass the arguments to the logging function and let it interpolate them.

ThrawnCA (changes requested):

ckanext/xloader/plugin.py
ThrawnCA: An external CSV can be loaded, yes, but then the format field should be CSV.

ckanext/xloader/plugin.py
ThrawnCA: https://docs.python.org/3/howto/logging.html#logging-variable-data

ckanext/xloader/plugin.py
JVickery-TBS: Oh like if the `datastore_active` is 0 or 1 instead of True/False?

ckanext/xloader/plugin.py
JVickery-TBS: Ohhh had no idea. Not exactly a python developer so still learning the pythonic stuff. Thanks! this is good to know. Logging calls have been updated to this (and did it in my ckanext-validation PR as well)

ckanext/xloader/plugin.py
ThrawnCA: No, the url_type can be `None` instead of an empty string.

JVickery-TBS: Okay this should be done now, the asbool helper method raise an error so had to deal with that

ckanext/xloader/plugin.py
```diff
@@ -233,9 +233,14 @@ def get_helpers(self):
...
         return False
+    has_url_type = True
+    try:
+        has_url_type = toolkit.asbool(res_dict.get('url_type'))
+    except ValueError:
+        pass
     return (not XLoaderFormats.is_it_an_xloader_format(res_dict.get('format', u''))
             and (res_dict.get('url_type') == 'upload'
-                or res_dict.get('url_type') == '')
+                or not has_url_type)
```
ThrawnCA, line 243: This should just use `not res_dict.get('url_type')`. `asbool` is not the right function for the job, see https://github.com/ckan/ckan/issues/7909

JVickery-TBS: Ah I see okay, Done!

JVickery-TBS:
@ThrawnCA `res['id']` changed to `resource_id` as well, and the `.lower()` removed from the debug logging on the format.

ThrawnCA (approved):

ThrawnCA:
@JVickery-TBS Conflicts need to be resolved after merging #198

JVickery-TBS:
@ThrawnCA conflicts resolved here

ckanext/xloader/plugin.py
```diff
@@ -108,7 +108,8 @@ def notify(self, entity, operation):
         We want to check if values have changed, namely the url and the format.
         See: ckan/model/modification.py.DomainObjectModificationExtension
         """
-        if operation != DomainObjectOperation.changed or not isinstance(entity, Resource):
+        if operation != DomainObjectOperation.changed \
+        or not isinstance(entity, Resource):
```
ThrawnCA, line 112: This needs to be indented for Flake8.

JVickery-TBS:
@ThrawnCA new tests are now passing as expected

duttonw (approved):

ThrawnCA (approved):

JVickery-TBS:
@ThrawnCA brought in ur test changes, thanks that looks a lot better and makes way way more sense haha.

Also merged master into this one to resolve conflicts

PR #197, opened 2023-11-08
Delete Datastore Table Button

JVickery-TBS:
feat(views): added view and button to delete ds table;

- Added view to delete datastore table.
- Added button to delete a resource's datastore table.

Adds a button to the Xloader data page to delete a Resource's datastore table.

### DIFF
--- a/ckanext/xloader/templates/xloader/resource_data.html
+++ b/ckanext/xloader/templates/xloader/resource_data.html
@@ -5,13 +5,17 @@
 {% block primary_content_inner %}
 
   {% set action = h.url_for('xloader.resource_data', id=pkg.name, resource_id=res.id) %}
+  {% set delete_action = h.url_for('xloader.delete_datastore_table', id=pkg.id, resource_id=res.id) %}
   {% set show_table = true %}
 
   {% block delete_ds_button %}
-    <form class="mb-3 d-inline-block">
+    <form method="post" action="{{ delete_action }}" class="mb-3 d-inline-block">
+      {{ h.csrf_input() if 'csrf_input' in h }}
       <a href="{{ h.url_for('xloader.delete_datastore_table', id=pkg.id, resource_id=res.id) }}"
         class="btn btn-danger pull-left"
+        type="submit"
         data-module="confirm-action"
+        data-module-with-data=true
         data-module-content="{{ _('Are you sure you want to delete this DataStore table and Data Dictionary?') }}"
         >{% block delete_datastore_button_text %}<i class="fa fa-remove"></i>{{ _('Delete DataStore table') }}{% endblock %}</a>
     </form>
--- a/ckanext/xloader/templates/xloader/resource_data.html
+++ b/ckanext/xloader/templates/xloader/resource_data.html
@@ -7,8 +7,18 @@
   {% set action = h.url_for('xloader.resource_data', id=pkg.name, resource_id=res.id) %}
   {% set show_table = true %}
 
+  {% block delete_ds_button %}
+    <form class="mb-3 d-inline-block">
+      <a href="{{ h.url_for('xloader.delete_datastore_table', id=pkg.id, resource_id=res.id) }}"
+        class="btn btn-danger pull-left"
+        data-module="confirm-action"
+        data-module-content="{{ _('Are you sure you want to delete this DataStore table and Data Dictionary?') }}"
+        >{% block delete_datastore_button_text %}<i class="fa fa-remove"></i>{{ _('Delete DataStore table') }}{% endblock %}</a>
+    </form>
+  {% endblock %}
+
   {% block upload_ds_button %}
-    <form method="post" action="{{ action }}" class="datapusher-form">
+    <form method="post" action="{{ action }}" class="datapusher-form mb-3 d-inline-block">
       {{ h.csrf_input() if 'csrf_input' in h }}
       <button class="btn btn-primary" name="save" type="submit">
         <i class="fa fa-cloud-upload"></i> {{ _('Upload to DataStore') }}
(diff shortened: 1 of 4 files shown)


ckanext/xloader/templates/xloader/resource_data.html
```diff
@@ -5,10 +5,24 @@
...
+  {% block delete_ds_button %}
+    <form method="post" action="{{ delete_action }}" class="mb-3 d-inline-block">
+      {{ h.csrf_input() if 'csrf_input' in h }}
+      <a href="{{ h.url_for('xloader.delete_datastore_table', id=pkg.id, resource_id=res.id) }}"
+        class="btn btn-danger pull-left"
+        type="submit"
+        data-module="confirm-action"
+        data-module-with-data=true
+        data-module-content="{{ _('Are you sure you want to delete this DataStore table and Data Dictionary?') }}"
+        >{% block delete_datastore_button_text %}<i class="fa fa-remove"></i>{{ _('Delete DataStore table') }}{% endblock %}</a>
```
ThrawnCA, line 20: Probably shouldn't include the word "table" here since it's not usually in the Datastore UI.

JVickery-TBS: Word "table" has been removed from user visible things here

JVickery-TBS:
@ThrawnCA for this PR, I also reverted the `ckanapi` dependency addition and used `get_action` from the toolkit instead

ThrawnCA (approved):

ThrawnCA:
Looks good from here.

ThrawnCA:
Ideally there should be a sanity check to ensure that the package ID is correct for the resource ID.

ThrawnCA (approved):

ckanext/xloader/templates/xloader/resource_data.html
```diff
@@ -5,10 +5,26 @@
...
   {% set action = h.url_for('xloader.resource_data', id=pkg.name, resource_id=res.id) %}
+  {% set delete_action = h.url_for('xloader.delete_datastore_table', id=pkg.id, resource_id=res.id) %}
   {% set show_table = true %}
 
+  {% block delete_ds_button %}
+    {% if res.datastore_active %}
+      <form method="post" action="{{ delete_action }}" class="mb-3 d-inline-block">
+        {{ h.csrf_input() if 'csrf_input' in h }}
+        <a href="{{ h.url_for('xloader.delete_datastore_table', id=pkg.id, resource_id=res.id) }}"
+          class="btn btn-danger pull-left"
```
ThrawnCA, line 16: This should use 'pull-right' so the upload button comes first.

ckanext/xloader/templates/xloader/resource_data.html
```diff
@@ -5,10 +5,26 @@
...
+          type="submit"
+          data-module="confirm-action"
+          data-module-with-data=true
+          data-module-content="{{ _('Are you sure you want to delete the DataStore and Data Dictionary?') }}"
+          >{% block delete_datastore_button_text %}<i class="fa fa-remove"></i>{{ _('Delete from DataStore') }}{% endblock %}</a>
```
ThrawnCA, line 21: Needs a space between the icon and text.

Also, consider the `fa-trash-alt` icon instead of `fa-remove`.

ThrawnCA (changes requested):
Testing on dev.data.qld.gov.au indicates that the positioning of the Delete button doesn't mesh well with the Upload to Datastore button.

ThrawnCA (changes requested):
After the confirm-action JavaScript runs, the button submits a datastore refresh instead of deletion.

Our fork will fix this with https://github.com/qld-gov-au/ckanext-xloader/pull/74

ckanext/xloader/templates/xloader/resource_data.html
JVickery-TBS: should stop using pulls in CKAN really. They truly are the worst thing for CSS layout and ordering like this. I can wrap in flex box and do a flex revers is probably best

JVickery-TBS: @ThrawnCA okay this should be good not

ckanext/xloader/templates/xloader/resource_data.html
JVickery-TBS: done

JVickery-TBS:
@ThrawnCA button alignment should be good now

ckanext/xloader/views.py
```diff
@@ -14,7 +14,15 @@ def get_blueprints():
 
 @xloader.route("/dataset/<id>/resource_data/<resource_id>", methods=("GET", "POST"))
 def resource_data(id, resource_id):
-    return utils.resource_data(id, resource_id)
+    rows = request.args.get('rows')
```
ThrawnCA, line 17: This is unrelated.

JVickery-TBS: I needed to merge master because of merge conflicts

ckanext/xloader/templates/xloader/resource_data.html
```diff
@@ -5,17 +5,35 @@
 {% block primary_content_inner %}
 
   {% set action = h.url_for('xloader.resource_data', id=pkg.name, resource_id=res.id) %}
+  {% set delete_action = h.url_for('xloader.delete_datastore_table', id=pkg.id, resource_id=res.id) %}
```
ThrawnCA, line 8: Testing at dev.data.qld.gov.au indicates that setting a variable at this point in the template is ineffective. It needs to be set inside the block where it's used.

The previous line inadvertently gets away with it because a blank action means "submit to the current URL", which is the desired behaviour.

ThrawnCA (changes requested):

ckanext/xloader/templates/xloader/resource_data.html
JVickery-TBS: @ThrawnCA cool, did not know about variables and blocks scopes.

This should be fixed now.

ThrawnCA (approved):

PR #200, opened 2023-11-24
Validation Extension Support

JVickery-TBS:
Adds support for the `ckanext-validation` plugin. Behind the new config option `ckanext.xloader.requires_validation`, Resources will be required to be validated before being xloadered.


ckanext/xloader/plugin.py
```diff
@@ -94,23 +97,54 @@ def configure(self, config_):
...
-                "id": resource.id,
+                "id": entity.id,
             },
         )
         self._submit_to_xloader(resource_dict)
 
     # IResourceController
 
     def after_resource_create(self, context, resource_dict):
+        if 'validation' in toolkit.config.get('ckan.plugins', []) and \
```
ThrawnCA, line 142: This will not catch variations on ckanext-validation that use different plugin definitions. Is it feasible to instead check for the presence of one of the ckanext-validation actions, like `resource_validation_show`?

ThrawnCA (changes requested):

ckanext/xloader/plugin.py
JVickery-TBS: @ThrawnCA ah makes sense, in the case that a developer subclasses the whole Validation Plugin Class? I have added the new check for the `resource_validation_show` action

ckanext/xloader/plugin.py
```diff
@@ -246,3 +246,11 @@ def get_helpers(self):
...
             "xloader_status_description": xloader_helpers.xloader_status_description,
         }
+
+
+def _is_validation_plugin_loaded():
+    try:
+        toolkit.get_action('resource_validation_show')
+    except KeyError:
+        return False
+    return True
```
ThrawnCA, line 260: This could be within the `try` block.

ckanext/xloader/plugin.py
```diff
@@ -212,3 +246,15 @@ def get_helpers(self):
             "xloader_status": xloader_helpers.xloader_status,
             "xloader_status_description": xloader_helpers.xloader_status_description,
         }
+
+
+def _is_validation_plugin_loaded():
+    """
+    Checks the existance of a logic action from the ckanext-validation
```
ThrawnCA, line 253: Misspelled "existence"

JVickery-TBS: fixed

ckanext/xloader/plugin.py
JVickery-TBS: return true now in try

ckanext/xloader/plugin.py
```diff
@@ -107,9 +107,23 @@ def notify(self, entity, operation):
...
+        or not isinstance(entity, Resource):
             return
+
+        # If the resource requires validation, stop here if validation
+        # has not been performed or did not succeed. The Validation
+        # extension will call resource_patch and this method should
+        # be called again. However, url_changed will not be in the entity
+        # once Validation does the patch.
+        if _is_validation_plugin_loaded() and \
+        toolkit.asbool(toolkit.config.get('ckanext.xloader.requires_validation')):
```
ThrawnCA, line 119: This needs indentation for Flake8.

ckanext/xloader/plugin.py
```diff
@@ -107,9 +107,23 @@ def notify(self, entity, operation):
...
+            if entity.__dict__.get('extras', {}).get('validation_status', None) != 'success':
```
ThrawnCA, line 120: What about resources without a schema? Even if the flag is set, they won't have a validation status, I think?

ThrawnCA (changes requested):

ckanext/xloader/plugin.py
JVickery-TBS: @ThrawnCA lemme know if it is good now on the flake8

ckanext/xloader/plugin.py
JVickery-TBS: same as above

JVickery-TBS:
@ThrawnCA I added the flash message to the `resource_data` POST request here too now. So will prevent users from using the "Upload to DataStore" button if the resource has not passed validation yet.

ckanext/xloader/plugin.py
JVickery-TBS: @ThrawnCA the default resource schema in core for `__extras` will accept anything. And the `ckanext-validation` plugin will patch resources with `validation_status`. So this should be fine unless if someone has modified the resource schema to not have extras and/or validation_status.

But I feel like if they have done that, then they are probably not using the validation extension.

ThrawnCA: We're running the validation plugin, but resources that don't specify any schema don't get a validation run or a `validation_status`. Eg https://dev.data.qld.gov.au/dataset/testing-qoldev-687/resource/de2f6122-1dee-4295-a589-8dfb702f5983 (https://dev.data.qld.gov.au/api/action/resource_show?id=de2f6122-1dee-4295-a589-8dfb702f5983)

ThrawnCA (changes requested):

ckanext/xloader/plugin.py
wardi: That's interesting. In our case we want all resources validated with a static schema (our users can't be trusted to provide one it seems) before being loaded.

Maybe we could support both cases with another config option, or by defining a helper function that determines whether a resource can be loaded that a theme plugin could then override?

JVickery-TBS:
@ThrawnCA fixed flake8 check here now as well.

ckanext/xloader/plugin.py
duttonw: In our situation. I do remember our business owner not wanting users who upload files to get annoyed if they provided csv's that were not valid even to basic rules and as such is skipped if schema is missing. 

Most annoying they did not do this as a flag and hard coded in our fork many many moons ago.
https://github.com/qld-gov-au/ckanext-validation/commit/0e42ce7dcd39bd6ba3cc17d94cf70a94b2c6a012#diff-9254636ee2e98ff1eaf809e6f490d1d2193f794c434d9e6e5235e3f2ccdfa717

JVickery-TBS: Hmmm @ThrawnCA @duttonw 

(Firstly my bad, I was thinking of the core code schemas not the `schema` field from the Validation plugin)

So I am not sure if the two config options would fully support your setup.

Mainly because even if I add in the second config option and check if the Resource's `schema` field is empty. You have this code (https://github.com/qld-gov-au/ckanext-validation/commit/0e42ce7dcd39bd6ba3cc17d94cf70a94b2c6a012#diff-35a1e13aaa4a67e7f831c29f02207acce46ed14129c73d3ecd1537bb9dcc89f9R77):
```
def _get_schema_or_default_schema(resource):
    schema = resource.get('schema')

    if asbool(resource.get('align_default_schema')):
        dataset = get_action('package_show')({}, {'id': resource['package_id']})

        if dataset:
            schema = dataset.get(u'default_data_schema')

    if schema and isinstance(schema, string_types):
        schema = schema if is_url_valid(schema) else json.loads(schema)

    return schema
```

So if a Resource does not have a `schema`, they can use its package's `default_data_schema`?

If that is the case, then if you have the new config option to require a validation report BUT then ignore empty Resource Schemas, it would then bypass this functionality of the package `default_data_schema`.

let me know your thoughts! Thanks!

JVickery-TBS: @ThrawnCA @duttonw okay sorry, I realize that code is only used in the badge helper.

https://github.com/qld-gov-au/ckanext-validation/blob/master/ckanext/validation/logic/action.py#L53
So in the `resource_validation_run` it just returns out if no `schema`.

Just to confirm: these no-schema resources still go to XLoader and DataStore??

duttonw:
Love the work you have done on this @JVickery-TBS .  One thing that I think would make this perfect is to see if xloader can hook into validation to run the xloader load in sync mode right after we have a success, or maybe another priority queue so its not delayed again. Since validation success ensure you don't have blank lines or duplicate headers, the fast load should always run and not defer back to messytables slow python orm style commits.

Just a bit of background:
Reason being is that in data.qld.gov.au we defer validation to the batch queue so that the user does not have a stalled process on large datasets or slow url collections.  We sometimes get alot of 1million row updated csv's loaded and if they fail the fast load take up to 1hour to be ingested. If we had a validation job occur and then had the update datastore put behind such a resource, then the author would get impatient and lodge a support ticket on why their resource is not updated.

Something to think about with this integration:
Another problem I forsee, mostly in the realm of the validation extension, is that when you save the dataset or re order said dataset, it may then re-validate ALL resources. It would be awesome if we could utilise part of xloader skip checks if the filehash is the same to save on duplicate runs.

Possible other extension enhancement for performance:
Unsure if your system uses the archiver extension. data.qld.gov.au does.  This means we have downloaded from url or off disk/blob store the file 3 times. or more due to validation loops. Thankfully the archiver is quite fast while the validation and xloader is quite slow.

ckanext/xloader/plugin.py
ThrawnCA: Sure, if it's a tabular format then XLoader will consume it, no schema required.

JVickery-TBS:
@duttonw Yeah we do the same thing with our setup as well in which we have to do async mode for validation and xloader as most of the CSVs that get upload are gigantic.

As for the UX for users waiting for these things, I am currently working on some UI things (https://github.com/open-data/ckanext-canada/pull/1430) to show the status of Xloader for a Resource, e.g. "Pending" or "Running" etc. Still kind of a work in progress, but let me know if this would be a good thing to put back into the Xloader repo.

Regarding the archiver, we actually use the cloud storage plugin, so things have to be gathered from there every time. But with the loop that happens, we kind of actually found out that this was an issue with the Core CKAN code and how the Resource plugin hooks work. Things tend to always get run multiple times because the Resource actions always call the Package actions. Still discussions to go on how to fix that. 

https://github.com/open-data/ckanext-validation/pull/9 <- I put in a temporary fix on our Validation fork to prevent the looping. Along with a temporary fix in our Xloader one (https://github.com/open-data/ckanext-xloader/commit/df5a4c0f5f0fc3bfcd4e4cf32e4370b86c9f6e81). These two things prevent any job looping. This was for sure a major issue as the resource_patch from Xloader would cause another Validation run, which would then cause another Xloader run and stop because file hash did not change, moreover, it would do that to ALL the resources in a dataset. So yeah, updating a single Resource should have enqueued 1 or 2 jobs, but could end up enqueuing like 70.

YES the validation and xloader were slow, so because of this loop, some datasets with 200+ resources in them backed up the job queue to a couple thousand. It was terrible and that is when I looked to fix this job looping thing. I will try and get my job looping prevention into the ckanext-validation repo this week. Even if this gets fixed sometime in upstream, should not really have a major impact with the work-around in Validation.

But as for your request to have validation do resource_patch and have Xloader ALWAYS do sync instead of async in that specific case, I shall look into that this week as well. Because, yeah, I feel like it could be annoying for a user to have to wait for Validation, and then after that passes, wait again for their Xloader job (that is now at the end of the queue) to run. Especially on busier sites or at busier times when the queue might be larger. (Especially with the job loop stuff happening, this is basically always hahaha)

ThrawnCA:
Another issue with asynchronous processing is that the logic in `plugin.py` to check whether XLoader can proceed will always see the _previous_ `validation_status`.

duttonw:
Hi @JVickery-TBS 

> As for the UX for users waiting for these things, I am currently working on some UI things (https://github.com/open-data/ckanext-canada/pull/1430) to show the status of Xloader for a Resource, e.g. "Pending" or "Running" etc. Still kind of a work in progress, but let me know if this would be a good thing to put back into the Xloader repo.

Please do create a new pr for the inclusion of the badge.

Place it behind a config option to disable badge (default is on)

When enabled:
Have these messages always visible:
```
 'pending': _('Data awaiting load to DataStore'),
 'running': _('Loading data into DataStore'),
 'error': _('Failed to load data into DataStore'),
```

Have another flag to enable all (debug level) messages:
```
 'complete': _('Data loaded into DataStore'),
 'active': _('Data available in DataStore'),
 'inactive': _('Resource not active in DataStore'),
 'unknown': _('DataStore status unknown'),
```
 
My reasoning for this is to reduce the clutter on resources. You only want to know when things are changing most of the time. Once its done its purpose is served.

I'd also only include the url for people who have author or above. Reason being is that we allow non-authors to register to provide 'data requests' as well as commenting and email notification on dataset/resources/groups etc (atm this is paused due to very bad sql performance on dashboard viewing which seem to be a recurring problem https://github.com/ckan/ckan/issues/6028 and https://github.com/ckan/ckan/pull/7878 )
```
pusher_url = t.h.url_for('xloader.resource_data',
                             id=resource.get('package_id'),
                             resource_id=resource.get('id'))
```

JVickery-TBS:
@duttonw okay sounds great I will make time in these coming weeks to do the above. (and yeaahhh the Following and the Dashboard stuff gets pretty slow, I think we have disabled following things and the dashboard as well).

@ThrawnCA Yeah, I was thinking of how to do the Validation Pass -> Sync Mode Xloader. I am just trying to figure out how to do it all in this plugin. As we have the button to "Upload to Datastore". So I don't know the logic to determine if the caller is from the View or from the Validation Job yet. (maybe I can see if we are currently in a job and see what the job callback is from the `ckan.lib.jobs`). We shall see!

duttonw:
Hi @JVickery-TBS , 

@ThrawnCA did something similar in our branch for the CLI click command for sys admins to force a datastore when under the crunch, you can see in in this pr https://github.com/qld-gov-au/ckanext-xloader/pull/82

Regards,

@duttonw

ThrawnCA:
> So I don't know the logic to determine if the caller is from the View or from the Validation Job yet. (maybe I can see if we are currently in a job and see what the job callback is from the `ckan.lib.jobs`). We shall see!

That could be easily fixed by putting a hidden input flag in the view. It wouldn't even be a problem if someone spotted it and chose to insert it into an API call.

JVickery-TBS:
@ThrawnCA @duttonw thanks for sending over that commit, I will check it out to see if I can use some of the code.

The issue that I am seeing here is now how RQ runs the jobs in workers. Firstly, you would have to have a worker running for that queue, and have it in the top priority. I would like to not have users need to run a separate queue. Have not read Thrawn's commit yet, but maybe it just executes the burst of that custom queue right away?

I would also like to support other devs in which they call the submit to xloader action, it should probably always enqueue by default.

I am currently having difficulties preventing loops when running the xloader directly from the validation job. So just trying to figure that one out.

JVickery-TBS:
@ThrawnCA @duttonw its not exactly where I want it, but am at a bit of a block right now.

I got it to a point in which if it is chaining from the Validation job, then it will enqueue a job and push it to the start of the queue.

Bad things with this are:

- the condition to check if we are inside of a validation job kinda sucks;
- validation can do sync mode itself, so does not necessarily even have a job;
- if a user has another queue other than default with a higher priority, then the push to top of default queue might not actually work great.

So I am not really sure at this point how to move forward, I feel like the best/easiest thing would be the add some feature to ckanext-validation. The extension DOES have a context variable called `_performed_validation`. This WOULD solve all the issues with the bad job check, and the sync validation style. But as it is inside of the context variable, the `notify` method from the `IDomainObjectModification` implementation does not have the context passed to it. So I am also trying to figure out a way to get that variable

ckanext/xloader/action.py
```diff
@@ -40,6 +40,10 @@ def xloader_submit(context, data_dict):
     :param ignore_hash: If set to True, the xloader will reload the file
         even if it haven't changed. (optional, default: False)
     :type ignore_hash: bool
+    :param sync: If set to True, the xloader callback will be executed right
```
ThrawnCA, line 43: I'm not comfortable with giving this power to all publishers. It would become incredibly easy to DoS the site by accident. Remember that synchronous processing will use the memory and CPU cycles of the web server.

It's different from providing a flag on the command line. That can only be used by staff who have top-level access and should be familiar with the technical capabilities of the system. This, though, is potentially open to non-technical users who only know that they're being pressured to make things happen faster.

If this feature is going to be present, it ought to require higher level authorisation such as a sysadmin.

ThrawnCA (changes requested):

ThrawnCA:
@JVickery-TBS Have you looked at the interaction between the archiver and QA plugins? Archiver defines an `IPipe` interface to receive callbacks, QA implements that interface, so when Archiver has finished uploading a resource, QA gets nudged to go and examine it.

A similar pattern could be used here, but would require changes to the Validation plugin to add callback support.

ckanext/xloader/action.py
JVickery-TBS: OMG I forgot that `xloader_submit` is a registered action. Yeah that is bad haha. Okay I will move the sync stuff around and see if I can figure out a bit more

JVickery-TBS:
requires: https://github.com/frictionlessdata/ckanext-validation/pull/97

ckanext/xloader/plugin.py
```diff
@@ -68,6 +76,23 @@ def configure(self, config_):
...
 
+    # IPipeValidation
+
+    def receive_validation_report(self, validation_report):
+        if utils.requires_successful_validation_report():
+            res_dict = toolkit.get_action('resource_show')({'ignore_auth': True},
+                                                           {'id': validation_report.get('resource_id')})
+            if (toolkit.asbool(toolkit.config.get('ckanext.xloader.validation.enforce_schema', True))
+                or res_dict.get('schema', None)) and validation_report.get('status') != 'success':
+                    # either validation.enforce_schema is turned on or it is off and there is a schema,
```
ThrawnCA, line 87: Comment could be clearer, but okay.

I might phrase it more as "A schema is either present, or required".

JVickery-TBS: Inline comments updated for clarity and human readability (I don't even know what I was trying to say now haha.)

ckanext/xloader/action.py
```diff
@@ -166,15 +173,20 @@ def xloader_submit(context, data_dict):
         job = enqueue_job(
             jobs.xloader_data_into_datastore, [data], queue=custom_queue,
             title="xloader_submit: package: {} resource: {}".format(resource_dict.get('package_id'), res_id),
-            rq_kwargs=dict(timeout=timeout)
+            rq_kwargs=dict(timeout=timeout, at_front=sync)
```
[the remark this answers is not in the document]
JVickery-TBS: @ThrawnCA Yes that would be an issue for sure, thanks for catching! I have added the `ignore_not_sysadmin` validator to the `sync` key in the schema. So only sysadmins and the system should be able to do it.

ckanext/xloader/action.py
```diff
@@ -166,15 +173,20 @@ def xloader_submit(context, data_dict):
...
         )
     except Exception:
-        log.exception('Unable to enqueued xloader res_id=%s', res_id)
+        if sync:
+            log.exception('Unable to xloader res_id=%s', res_id)
+        else:
+            log.exception('Unable to enqueued xloader res_id=%s', res_id)
```
ThrawnCA, line 182: "Unable to enqueued" -> "Unable to enqueue"

ckanext/xloader/schema.py
```diff
@@ -16,6 +16,7 @@
 boolean_validator = get_validator('boolean_validator')
 int_validator = get_validator('int_validator')
 OneOf = get_validator('OneOf')
+ignore_not_sysadmin = get_validator('ignore_not_sysadmin')
```
ThrawnCA, line 19: That's quite helpful to know about, thank you.

ThrawnCA (approved):

duttonw (approved):
Other than @ThrawnCA  debug wording tweak. It looks good. 

Is these commits used anywhere else (i.e. is it a candidate for squashing for clean history).

Can you also update the README.md and maybe the Change log of what you added. 

Since I've recently become a mantainer of the pypi module for this so I'd like to push this out to that soon after.  (Doing it correctly with a github action pipeline)

JVickery-TBS:
@duttonw @ThrawnCA squashed version: https://github.com/ckan/ckanext-xloader/pull/237

PR #209, opened 2024-02-28
In plugin.py, there is an fix of resource format key error

Nisha1293:
This is the fix of resource format key error

### DIFF
--- a/ckanext/xloader/plugin.py
+++ b/ckanext/xloader/plugin.py
@@ -148,11 +148,12 @@ class xloaderPlugin(plugins.SingletonPlugin):
 
     def _submit_to_xloader(self, resource_dict):
         context = {"ignore_auth": True, "defer_commit": True}
-        if not XLoaderFormats.is_it_an_xloader_format(resource_dict["format"]):
+        resource_format = resource_dict.get("format")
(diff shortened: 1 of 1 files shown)


Nisha1293:
Hi @duttonw 
I have raised one PR.  Could you please review it.

duttonw:
Hi @Nisha1293,

Can you enable github actions on your repo and retrigger the build so i can review the tests etc.
https://github.com/Nisha1293/ckanext-xloader/actions

regards,

@duttonw

ThrawnCA:
I don't understand what situation would fail to have a format.

duttonw:
I think its in relation to not having 'format' in the dict. Some ckan admin's may have decided to drop that field on their servers.

ThrawnCA:
> Some ckan admin's may have decided to drop that field on their servers.

And still run XLoader, which relies on the format field to know whether it should try to parse a resource? That seems very odd indeed.

duttonw:
@ThrawnCA, I don't see any issue in merging this since its just wrapped it in a null safe context. 

Your thoughts?

ThrawnCA:
> Your thoughts?

I suppose it's harmless, it just seems very odd for anyone to need this null-check.

duttonw (approved):

Issue #202, opened 2024-01-03
Add synchronous CLI option


ThrawnCA:
It would sometimes be beneficial to have the ability to immediately and synchronously run an XLoader update from the command line, instead of submitting a job to the queue (which may be long). Rather like how ckanext-archiver has both `update` to submit an asynchronous job, or `update-test` to run immediately.

KatiRG:
@ThrawnCA Is it possible to run xloader synchronously in the code itself?

duttonw:
The simple answer is yes, but at what cost. On our ckan instance you may be waiting up to 40 minutes to process a single file. That would mean the publisher doing the request would be left hanging or the cdn in front would have cut the connection after 60-300sec of wait time. 


The front ends goal is to be as fast as possible which pushes for said need to decouple. It may be possible to notify triggers instead of a queue but it's still putting load on that single instance instead of across the front end farm or dedicated loading server. 

When you have the harvester plugin installed it requests you to make two extra queues. The general idea is you could run direct via cli or add to a less busy queue to get priority loading.

KatiRG:
Yes, that is the cost. Unfortunately, we are developing a data validation pipeline that requires the dataset to already be in the datastore before it can continue. So we have to wait for xloader to finish the job. Is there already a function that can wait for xloader to complete the submission? Or at least, a function that can continually poll the task status and terminate when the status is complete?

duttonw:
Based on your requirements, do create a new extension that hooks into xloader finish and triggers your pipeline via https.
You may want to make a sysadmin screen to view/alter end points and I do feel building for a popular spec like [aws sqs](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessage.html) for the payload would be best as well as maybe the ability to fire different endpoints based on the resource uuid/group/org or dataset tags.

Please see https://github.com/ckan/ckanext-xloader/blob/master/ckanext/xloader/action.py#L270 where there is already a hook system already available for you to connect into when the loader is complete || table loaded but index's still being built. 
```
 for plugin in p.PluginImplementations(xloader_interfaces.IXloader):
            plugin.after_upload(context, resource_dict, dataset_dict)
```

Looking over https://data.ontario.ca/api/3/action/status_show vs https://www.data.qld.gov.au/api/3/action/status_show

[...]

PR #212, opened 2024-03-13
add more options for maintainers to expedite XLoader runs, GitHub #202

ThrawnCA:
- Add CLI flag for loading immediately instead of queueing
- Allow sysadmins to specify an alternate queue to run on

### DIFF
--- a/ckanext/xloader/action.py
+++ b/ckanext/xloader/action.py
@@ -46,12 +46,12 @@ def xloader_submit(context, data_dict):
 
     :rtype: bool
     '''
+    p.toolkit.check_access('xloader_submit', context, data_dict)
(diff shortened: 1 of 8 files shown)


duttonw (approved):

EricSoroos:
I don't run xloader anywhere -- what's the point of the alternate queue?

duttonw:
It's very useful When you have a couple of 1.5 million record files which fail to fast load and take 40mins to slow load making the default queue up to 4 hours deep. 

This allows a sysadmin to force a priority queue load for a dataset if an author/publisher requests it. 

In data.qld.gov.au use case this occurs when the weekly Covid stats are updated and a couple of other datasets.

ThrawnCA:
Note that the alternate queue is not currently surfaced in the UI, but could be specified in an API call.

The synchronous CLI run is also useful for things like testing.

PR #219, opened 2024-07-01
Rewrite site url

stojanovskis1:


duttonw:
Hi @stojanovskis1 ,

Thanks for wanting to contribute back to xloader extension. 


Could you please split this into two branches/PR's:
* one for  the xlsx -> csv pre-converter. 
  
  Please add a config flag to disable this conversion so it follows the legacy path (also defensively code it that if the library is not available that it falls back to legacy processing)

* Rewrite site url: What is the business use case for doing the site url re-write with CKANEXT__XLOADER__REWRITE_SITE_URL or internal config name?



Please also add unit tests on the edge cases for each of the Pull Requests, i.e. an example xls/xlsx file that you want to process that was not processing correctly in the legacy path. As well as unit tests on the site url rename and a path testing its not changed when not set, etc.

As for each PR, please also update the README on the new config/os flag's and what they do.

Regards,

@duttonw

PR #220, opened 2024-07-09
CKAN 2.11 support

amercader:


amercader:
The [failure on 2.11](https://github.com/ckan/ckanext-xloader/actions/runs/9855271725/job/27209839735?pr=220) looks interesting. Seems like there is an issue with the field types being created:

```
>       assert self._get_column_types(Session, resource_id) == [
            u"int4",                                                 
            u"tsvector",                                             
            u"timestamp",                                            
            u"numeric",                                              
            u"text",                                                 
        ]                                                            
E       AssertionError: assert ['int4', 'tsvector', 'text', 'text', 'text'] == ['int4', 'tsvector', 'timestamp', 'numeric', 'text']
E                                                                    
E         At index 2 diff: 'text' != 'timestamp'
E                                                                    
E         Full diff:                                                 
E           [                                                        
E               'int4',                                              
E               'tsvector',                                          
E         -     'timestamp',                                         
E         -     'numeric',                                           
E         +     'text',                                              
E         +     'text',                                              
E               'text',                                              
E           ]                                                        

ckanext/xloader/tests/test_loader.py:763: AssertionError
```
@wardi can you think of a change in the datastore since 2.10 that might cause this? Perhaps the field info stuff related to Table Designer?

wardi:
Interesting, maybe IDataDictionaryForm-related if xloader can't find its type overrides. I'll take a look

PR #223, opened 2024-07-31
RQ Job Timeout Handling

JVickery-TBS:
fix(logic): close temp file on rq timeout;

- Always close the temp file on exception handling.
- Added handling for the rq job timeout.


ckanext/xloader/jobs.py
```diff
@@ -222,6 +222,11 @@ def tabulator_load():
                               resource_id=resource['id'],
                               mimetype=resource.get('format'),
                               logger=logger)
+        except rq_timeouts.JobTimeoutException as e:
```
ThrawnCA, line 225: Wouldn't it be better to put this further out, eg line 256?

JVickery-TBS: @ThrawnCA moved out to wrap the `tabulator_load` method call with the try/catch. Lemme know if that is actually the right place. Thanks!

ckanext/xloader/jobs.py
```diff
@@ -245,10 +245,21 @@ def tabulator_load():
     logger.info("'use_type_guessing' mode is: %s", use_type_guessing)
     try:
         if use_type_guessing:
-            tabulator_load()
+            try:
+                tabulator_load()
+            except rq_timeouts.JobTimeoutException as e:
```
ThrawnCA, line 250: This isn't catching timeouts from the tabulator_load on line 266. Perhaps move this block to the outer catch, line 267?

ckanext/xloader/jobs.py
ThrawnCA: It's a start, but I think line 267 would be a better place.

duttonw:
@JVickery-TBS  i'd like the idea but with the tests failing, I can't look at merging this. Also unsure about what @ThrawnCA picked up on another path that is not caught correctly.

duttonw:
Hi @JVickery-TBS ,

I'd like to incorperate this into xloader. Are you able to rebase to master so that we have the tests fixed.

Regards,

@duttonw

ckanext/xloader/jobs.py
JVickery-TBS: moved to outer try/catch alongside the `FileCouldNotBeLoadedError` one.

ckanext/xloader/jobs.py
```diff
@@ -259,6 +260,14 @@ def tabulator_load():
                 logger.warning('Load using COPY failed: %s', e)
                 logger.info('Trying again with tabulator')
                 tabulator_load()
+    except JobTimeoutException as e:
+        try:
+            tmp_file.close()
+        except FileNotFoundError:
+            pass
+        timeout = config.get('ckanext.xloader.job_timeout', '3600')
```
duttonw, line 268: can you use RETRIED_JOB_TIMEOUT as its the same variable load.

ckanext/xloader/tests/test_jobs.py
```diff
@@ -123,6 +137,16 @@ def test_data_max_excerpt_lines_config(self, cli, data):
         resource = helpers.call_action("resource_show", id=data["metadata"]["resource_id"])
         assert resource["datastore_contains_all_records_of_source_file"] is False
 
+    def test_data_with_rq_job_timeout(self, cli, data):
+        file_suffix = 'multiplication_2.csv'
+        self.enqueue(jobs.xloader_data_into_datastore, [data], rq_kwargs=dict(timeout=15))
```
duttonw, line 142: could set to 2seconds so testing is faster.

JVickery-TBS:
@duttonw lemme know if this one is good, and I can make a squashed version

JVickery-TBS:
@duttonw @ThrawnCA squashed version: https://github.com/ckan/ckanext-xloader/pull/239 (and added to the changelog file too)

ckanext/xloader/jobs.py
```diff
@@ -259,6 +260,13 @@ def tabulator_load():
                 logger.warning('Load using COPY failed: %s', e)
                 logger.info('Trying again with tabulator')
                 tabulator_load()
+    except JobTimeoutException as e:
+        try:
+            tmp_file.close()
+        except FileNotFoundError:
+            pass
+        logger.warning('Job timed out after %ss', RETRIED_JOB_TIMEOUT)
+        raise JobError('Job timed out after {}s'.format(RETRIED_JOB_TIMEOUT))
```
ThrawnCA, line 269: If you're generating this string anyway, might as well use it for the logger as well.

PR #225, opened 2024-09-10
SQLAlchemy v2 support

smotornyuk:
Compatibility with the new SQLAlchemy version tested against [this PR](https://github.com/ckan/ckan/pull/8410).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -53,8 +53,6 @@ jobs:
             ckan-image: "2.11-py3.10"
           - ckan-version: "2.10"
             ckan-image: "2.10-py3.10"
-          - ckan-version: "2.9"
-            ckan-image: "2.9-py3.9"
           #- ckan-version: "master" Publish does not care about master
           #  ckan-image: "master"
       fail-fast: false
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -30,9 +30,6 @@ jobs:
           - ckan-version: "2.10"
             ckan-image: "2.10-py3.10"
             experimental: false
-          - ckan-version: "2.9"
-            ckan-image: "2.9-py3.9"
-            experimental: false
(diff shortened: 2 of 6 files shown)


duttonw:
Looks like setup tools for 2.9 has started to fail unrelated to your changes. 


      /usr/lib/python3.9/site-packages/setuptools/_distutils/dist.py:261: UserWarning: Unknown distribution option: 'tests_require'

ThrawnCA:
> Looks like setup tools for 2.9 has started to fail unrelated to your changes.

We can fix that, but I'll note that with the release of CKAN 2.11, 2.9 is no longer officially supported: https://ckan.org/blog/the-new-ckan-211-release-is-now-available

ThrawnCA:
Looks like 2.9 isn't compatible with the newer SQLAlchemy. We should probably make version-specific requirements files.

duttonw:
I've cut a release today for 1.2.0 which is compatible for 2.9 ~ 2.11. 

I'm ok with dropping 2.9 support via this PR, I've updated the doc's and cicd pipeline accordingly. 

I'd like to pull in the last pieces of work from the qld-gov-au fork prior to this being included for one last 1.x.x release.

As this is major breaking change, it should bump to 2.0.0 on this being merged into master/released.

ThrawnCA (approved):

ThrawnCA (approved):

.github/workflows/publish.yml
```diff
@@ -53,8 +53,6 @@ jobs:
             ckan-image: "2.11-py3.10"
           - ckan-version: "2.10"
             ckan-image: "2.10-py3.10"
-          - ckan-version: "2.9"
```
ThrawnCA, line 56: Still needs to be removed from the test workflow.

ThrawnCA (changes requested):

ThrawnCA (approved):

PR #231, opened 2024-10-25
Fix automated PyPI publishing

ThrawnCA:
- Remove unwanted 'v' from tag pattern
- Fix environment parameter to have a default value when automatically triggered
- Add validation that tag version matches Python version


duttonw:
[Invalid workflow file: .github/workflows/publish.yml#L7](https://github.com/ckan/ckanext-xloader/actions/runs/11509720051/workflow)
You have an error in your yaml syntax on line 7

duttonw (approved):

PR #234, opened 2024-11-18
Proof of concept for new config option: ckanext.xloader.site_url

pwalsh:
We want to be able to communicate within a docker network without using the public site_url. This is a minimal proof of concept for achieving this.

Update: There are 2 scenario's where XLoader communicates to CKAN

**Scenario 1:** When it downloads the resource file from CKAN ie: in `_download_resource_data()`
**Scenario 2:** When it updates the status of the XLoader job ie: in `callback_xloader_hook()`

If the config option `ckanext.xloader.site_url` is set then **all** communication from XLoader to CKAN will use this URL. In scenario 1, if it's not set then the `ckan.site_url` config option will be used **unless** the download ie: `_download_resource_data()` `original_url` is different to `ckan_url` (ie: `ckan.site_url`) and in that case the `_download_resource_data()` download will use the `original_url` but all other communication back to CKAN (ie: scenario 2) will use the `ckan.site_url` option

### DIFF
--- a/ckanext/xloader/action.py
+++ b/ckanext/xloader/action.py
@@ -146,7 +146,7 @@ def xloader_submit(context, data_dict):
         "metadata": {
             "ignore_hash": data_dict.get("ignore_hash", False),
             "ckan_url": config.get("ckanext.xloader.site_url")
-            or config["ckan.site_url"],
+                or config["ckan.site_url"],
             "resource_id": res_id,
             "set_url_type": data_dict.get("set_url_type", False),
             "task_created": task["last_updated"],
--- a/ckanext/xloader/action.py
+++ b/ckanext/xloader/action.py
@@ -140,17 +140,18 @@ def xloader_submit(context, data_dict):
         qualified=True
     )
     data = {
-        'api_key': utils.get_xloader_user_apitoken(),
-        'job_type': 'xloader_to_datastore',
-        'result_url': callback_url,
-        'metadata': {
-            'ignore_hash': data_dict.get('ignore_hash', False),
-            'ckan_url': config['ckan.site_url'],
-            'resource_id': res_id,
-            'set_url_type': data_dict.get('set_url_type', False),
-            'task_created': task['last_updated'],
-            'original_url': resource_dict.get('url'),
-        }
+        "api_key": utils.get_xloader_user_apitoken(),
+        "job_type": "xloader_to_datastore",
+        "result_url": callback_url,
+        "metadata": {
+            "ignore_hash": data_dict.get("ignore_hash", False),
+            "ckan_url": config.get("ckanext.xloader.site_url")
+            or config["ckan.site_url"],
+            "resource_id": res_id,
+            "set_url_type": data_dict.get("set_url_type", False),
+            "task_created": task["last_updated"],
+            "original_url": resource_dict.get("url"),
+        },
     }
     if custom_queue != rq_jobs.DEFAULT_QUEUE_NAME:
         # Don't automatically retry if it's a custom run
--- a/ckanext/xloader/config_declaration.yaml
+++ b/ckanext/xloader/config_declaration.yaml
@@ -2,6 +2,12 @@ version: 1
 groups:
   - annotation: ckanext-xloader settings
     options:
+      - key: ckanext.xloader.site_url
+        default:
+        description: |
+            Provide an alternate site URL for the xloader_submit action.
+            This is useful, for example, when the site is running within a docker network.
+        required: false
       - key: ckanext.xloader.jobs_db.uri
         default: sqlite:////tmp/xloader_jobs.db
         description: |
@@ -152,5 +158,3 @@ groups:
           they will also display "complete", "active", "inactive", and "unknown".
         type: bool
         required: false
-
-
--- a/ckanext/xloader/config_declaration.yaml
+++ b/ckanext/xloader/config_declaration.yaml
@@ -3,10 +3,12 @@ groups:
   - annotation: ckanext-xloader settings
     options:
       - key: ckanext.xloader.site_url
-        default:
+        example: http://ckan-dev:5000
+        default: http://ckan-dev:5000
         description: |
             Provide an alternate site URL for the xloader_submit action.
             This is useful, for example, when the site is running within a docker network.
+        validators: configured_default("ckan.site_url",None)
         required: false
       - key: ckanext.xloader.jobs_db.uri
         default: sqlite:////tmp/xloader_jobs.db
--- a/ckanext/xloader/config_declaration.yaml
+++ b/ckanext/xloader/config_declaration.yaml
@@ -4,7 +4,7 @@ groups:
     options:
       - key: ckanext.xloader.site_url
         example: http://ckan-dev:5000
-        default: http://ckan-dev:5000
+        default:
         description: |
             Provide an alternate site URL for the xloader_submit action.
             This is useful, for example, when the site is running within a docker network.
--- a/ckanext/xloader/jobs.py
+++ b/ckanext/xloader/jobs.py
@@ -212,7 +212,11 @@ def xloader_data_into_datastore_(input, job_dict, logger):
         set_datastore_active(data, resource, logger)
         if 'result_url' in input:
             job_dict['status'] = 'running_but_viewable'
-            callback_xloader_hook(result_url=input['result_url'],
+            callback_url = config.get('ckanext.xloader.site_url') or config.get('ckan.site_url')
+            callback_url = urljoin(
+                callback_url.rstrip('/'), '/api/3/action/xloader_hook')
+            result_url = callback_url
+            callback_xloader_hook(result_url=result_url,
                                   api_key=api_key,
                                   job_dict=job_dict)
         logger.info('Data now available to users: %s', resource_ckan_url)
--- a/ckanext/xloader/jobs.py
+++ b/ckanext/xloader/jobs.py
@@ -24,7 +24,7 @@ from ckan.plugins.toolkit import get_action, asbool, enqueue_job, ObjectNotFound
 
 from . import db, loader
 from .job_exceptions import JobError, HTTPError, DataTooBigError, FileCouldNotBeLoadedError
-from .utils import datastore_resource_exists, set_resource_metadata, get_ckan_url
+from .utils import datastore_resource_exists, set_resource_metadata, modify_resource_url, modify_ckan_url
 
 try:
     from ckan.lib.api_token import get_user_from_token
@@ -82,13 +82,10 @@ def xloader_data_into_datastore(input):
     # stillborn, for when xloader_submit is deciding whether another job would
     # be a duplicate or not
 
-    callback_url = get_ckan_url()
-    callback_url = urljoin(
-        callback_url.rstrip('/'), '/api/3/action/xloader_hook')
-
     job_dict = dict(metadata=input['metadata'],
                     status='running')
-    callback_xloader_hook(result_url=callback_url,
+
+    callback_xloader_hook(result_url=input['result_url'],
                           api_key=input['api_key'],
                           job_dict=job_dict)
 
@@ -150,7 +147,7 @@ def xloader_data_into_datastore(input):
         errored = True
     finally:
         # job_dict is defined in xloader_hook's docstring
-        is_saved_ok = callback_xloader_hook(result_url=callback_url,
+        is_saved_ok = callback_xloader_hook(result_url=input['result_url'],
                                             api_key=input['api_key'],
                                             job_dict=job_dict)
         errored = errored or not is_saved_ok
@@ -211,10 +208,7 @@ def xloader_data_into_datastore_(input, job_dict, logger):
         set_datastore_active(data, resource, logger)
         if 'result_url' in input:
             job_dict['status'] = 'running_but_viewable'
-            callback_url = get_ckan_url()
-            callback_url = urljoin(
-                callback_url.rstrip('/'), '/api/3/action/xloader_hook')
-            callback_xloader_hook(result_url=callback_url,
+            callback_xloader_hook(result_url=input['result_url'],
                                   api_key=api_key,
                                   job_dict=job_dict)
         logger.info('Data now available to users: %s', resource_ckan_url)
@@ -286,21 +280,16 @@ def _download_resource_data(resource, data, api_key, logger):
     data['datastore_contains_all_records_of_source_file'] = False
     which will be saved to the resource later on.
     '''
-    # check scheme
+ 
     url = resource.get('url')
-    url_parts = urlsplit(url)
+    url = modify_resource_url(url)
+    # check scheme
+    url_parts = urlsplit(url) 
     scheme = url_parts.scheme
     if scheme not in ('http', 'https', 'ftp'):
         raise JobError(
             'Only http, https, and ftp resources may be fetched.'
         )
-
-    resource_uri = urlunsplit(('', '', url_parts.path, url_parts.query, url_parts.fragment))
-    callback_url = get_ckan_url()
-    url = urljoin(
-        callback_url.rstrip('/'), resource_uri)
-    
-    url_parts = urlsplit(url) # reparse the url after the callback_url is set
 
     # fetch the resource data
     logger.info('Fetching from: {0}'.format(url))
@@ -455,7 +444,8 @@ def callback_xloader_hook(result_url, api_key, job_dict):
         else:
             header, key = 'Authorization', api_key
         headers[header] = key
-
+  
+    result_url = modify_ckan_url(result_url, job_dict['metadata']['ckan_url'])
     try:
         result = requests.post(
             result_url,
--- a/ckanext/xloader/jobs.py
+++ b/ckanext/xloader/jobs.py
@@ -12,10 +12,12 @@ import traceback
 import sys
 
 from psycopg2 import errors
-from six.moves.urllib.parse import urlsplit
+from six.moves.urllib.parse import urlsplit, urlparse, urlunparse
 import requests
 from rq import get_current_job
 import sqlalchemy as sa
+
+from urllib.parse import urljoin, urlunsplit
 
 from ckan import model
 from ckan.plugins.toolkit import get_action, asbool, enqueue_job, ObjectNotFound, config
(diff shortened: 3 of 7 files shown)


ThrawnCA:
CKAN 2.11 build is broken by https://github.com/ckan/ckan-docker-base/issues/86 and needs to run the test container as `root` to work around it.

ckanext/xloader/config_declaration.yaml
```diff
@@ -2,6 +2,12 @@ version: 1
 groups:
   - annotation: ckanext-xloader settings
     options:
+      - key: ckanext.xloader.site_url
+        default:
+        description: |
+            Provide an alternate site URL for the xloader_submit action.
+            This is useful, for example, when the site is running within a docker network.
```
wardi, line 9:
```suggestion
            This is useful, for example, when the site is running within a docker network
            or where the job runner can't access ckan using its normal public site URL.
```

ckanext/xloader/config_declaration.yaml
```diff
@@ -2,6 +2,12 @@ version: 1
 groups:
   - annotation: ckanext-xloader settings
     options:
+      - key: ckanext.xloader.site_url
+        default:
```
wardi, line 6: I think we don't need an empty string default, it's normal for optional settings to be not present if not provided
```suggestion
```

ckanext/xloader/plugin.py
```diff
@@ -61,13 +61,11 @@ def configure(self, config_):
...
-                raise Exception(
-                    "Config option `{0}` must be set to use ckanext-xloader.".format(
-                        config_option
-                    )
-                )
+        site_url_configs = ("ckan.site_url", "ckanext.xloader.site_url")
+        if not any(site_url_configs):
+            raise Exception(
+                f"One of config options {site_url_configs} must be set to use ckanext-xloader."
+            )
```
wardi, line 68: ckan will refuse to start if a site_url isn't provided, so this code would never get executed

ckanext/xloader/config_declaration.yaml
wardi: instead of the `toolkit.config.get("ckanext.xloader.site_url") or toolkit.config.get("ckan.site_url")` logic below we should be able to use just `toolkit.config.get("ckanext.xloader.site_url")` along with:
```
       validators: configured_default("ckan.site_url",None)
```

ckanext/xloader/plugin.py
pwalsh: @wardi here, I'm just extending the existing check for that. I prefer not to remove existing behavior, but just to provide the minimal new behavior required for the PR

https://github.com/ckan/ckanext-xloader/blob/master/ckanext/xloader/plugin.py#L64

ckanext/xloader/config_declaration.yaml
pwalsh: @wardi ok I see. so that default is just set in the config yaml, and then, any code can assume that fallback default for the `ckanext.xloader.site_url` setting?

wardi: yeah, this validator will set `ckanext.xloader.site_url` to the same as `ckan.site_url` when it's not given. Another one of @smotornyuk 's clever ideas.

ckanext/xloader/plugin.py
duttonw: If you did not cleanup the error message, this would not have been commented upon ;) 

I'm fine leaving this in as its a belts and braces approach, better to fail early (if it does occur which is very very remote)

ckanext/xloader/tests/test_jobs.py
```diff
@@ -59,17 +59,18 @@ def data(create_with_upload, apikey):
...
-            'set_url_type': False,
-            'task_created': datetime.utcnow().isoformat(),
-            'original_url': resource["url"],
-        }
+        "api_key": apikey,
+        "job_type": "xloader_to_datastore",
+        "result_url": callback_url,
+        "metadata": {
+            "ignore_hash": True,
+            "ckan_url": toolkit.config.get("ckanext.xloader.site_url")
```
duttonw, line 67: going to need more tests created to set these two values for proper validation.
one for default site url, another for the xloader override value.  updating the fixture is so/so.

duttonw (changes requested):
would like to see more tests to ensure all paths are covered.

pwalsh:
@duttonw yeah sure, will add tests, this is in draft as it is _not_ ready for review by maintainers, I'm just doing a minimal validation of the approach ( I work with @wardi and @amercader at Link Digital ).

JVickery-TBS:
Does this need to also handle the download of the temp file in `_download_resource_data` method of `ckanext.xloader.jobs`?

ckanext/xloader/config_declaration.yaml
```diff
@@ -3,10 +3,12 @@ groups:
   - annotation: ckanext-xloader settings
     options:
       - key: ckanext.xloader.site_url
-        default:
+        example: http://ckan-dev:5000
+        default: http://ckan-dev:5000
```
ThrawnCA, line 7: We should leave this blank if the value isn't provided.

ckanext/xloader/jobs.py
```diff
@@ -12,11 +12,13 @@
 import sys
 
 from psycopg2 import errors
-from six.moves.urllib.parse import urlsplit
+from six.moves.urllib.parse import urlsplit, urlparse, urlunparse
```
ThrawnCA, line 15: `six` is redundant since we no longer support Python 2.

ThrawnCA (changes requested):

ckanext/xloader/config_declaration.yaml
ThrawnCA: Do we need to bother with that when `jobs.py` already has the fallback `config.get('ckanext.xloader.site_url') or config.get('ckan.site_url')`?

IMO we should keep the fallback in `jobs.py` in case someone manually sets `ckanext.xloader.site_url` to a blank value (which could happen if eg the config is being automatically generated/populated from somewhere, perhaps using `ckanext-ssm-config`), so the validator approach is redundant.

ckanext/xloader/jobs.py
```diff
@@ -212,7 +212,11 @@ def direct_load():
         set_datastore_active(data, resource, logger)
         if 'result_url' in input:
             job_dict['status'] = 'running_but_viewable'
-            callback_xloader_hook(result_url=input['result_url'],
+            callback_url = config.get('ckanext.xloader.site_url') or config.get('ckan.site_url')
+            callback_url = urljoin(
+                callback_url.rstrip('/'), '/api/3/action/xloader_hook')
+            result_url = callback_url
```
ThrawnCA, line 218: Why create `result_url` at all?

kowh-ai: You mean assign `callback_url` back to `result_url` ? I'm try to minimise changes to the rest of the XLoader code. If you think I should just use `callback_url` then I can do that

ThrawnCA: You're already changing from `result_url=input['result_url']` to `result_url=result_url`. Why not just change it to `result_url=callback_url`? Or just call it `result_url` instead of `callback_url` in the first place.

ckanext/xloader/jobs.py
```diff
@@ -285,6 +297,14 @@ def _download_resource_data(resource, data, api_key, logger):
             'Only http, https, and ftp resources may be fetched.'
         )
 
+    resource_uri = urlunsplit(('', '', url_parts.path, url_parts.query, url_parts.fragment))
+    callback_url = config.get('ckanext.xloader.site_url') or config.get('ckan.site_url')
+    callback_url = urljoin(
+        callback_url.rstrip('/'), resource_uri)
+    
+    url = callback_url
+    url_parts = urlsplit(url) # reparse the url after the callback_url is set
```
duttonw, line 306: This will break all 'external' url's which are not internally referenced.  We should be using a 'generic' function that takes a url and does the swap if the original url.

Details below on my reasoning:

I see in my jobs stats table items is

```
{"metadata": {"ignore_hash": false, "ckan_url": "https://www.data.qld.gov.au", "resource_id": "5d1e8368-7ec3-435a-92d0-280ad1e3db0d", "set_url_type": false, "task_created": "2025-01-11 00:15:03.218182", "original_url": "https://apps.des.qld.gov.au/data-sets/water/gregory-environmental.csv?timestamp=2025-01-10T10:15:02+10:00", "datastore_contains_all_records_of_source_file": true, "datastore_active": true}, "status": "complete"}
```
using this as an example. The input of the _download_resource_data function is 
``res_dict = get_action('resource_show')(context, {'id': resource_id})``

looking at : https://www.data.qld.gov.au/api/action/resource_show?id=5d1e8368-7ec3-435a-92d0-280ad1e3db0d

on line 292, it gets the url which would be:

"url": "https://apps.des.qld.gov.au/data-sets/water/gregory-environmental.csv?timestamp=2025-01-16T10:15:01+10:00",

Which since you are 'force' changing the url, instead of checking its pointing to the CKAN site url first. it would then end up with 

http://localhost:5000/data-sets/water/gregory-environmental.csv?timestamp=2025-01-16T10:15:01+10:00
if the ckanext.xloader.site_url is http://localhost:5000

which would then be a 404 or hanging situation.

ckanext/xloader/tests/test_jobs.py
```diff
@@ -59,17 +59,18 @@ def data(create_with_upload, apikey):
...
-        'result_url': callback_url,
-        'metadata': {
-            'ignore_hash': True,
-            'ckan_url': toolkit.config.get('ckan.site_url'),
-            'resource_id': resource["id"],
-            'set_url_type': False,
-            'task_created': datetime.utcnow().isoformat(),
-            'original_url': resource["url"],
-        }
+        "api_key": apikey,
```
duttonw, line 62: please also add a test for 'external' url which is not using the site_url for the file.

duttonw (changes requested):
As it is, it will break external urls for csv files. I also suggest creating a util function that does the url swapping on a per url basis if it is using the 'site_url' that is in our config's. (so it does not break external urls).

ckanext/xloader/jobs.py
kowh-ai: Thanks @duttonw for taking the time to help me with this. It is much appreaciated.

Yes, I agree that a **util/helper function** is needed to swap out the CKAN URL

With regards to your scenario, why would you set `ckanext.xloader.site_url` to something if it broke communication? The `ckanext.xloader.site_url` config option does **not** exist at the moment in master. If `ckanext.xloader.site_url` is **not set** then the `callback_url` would just fall back to `ckan.site_url` which is mandatory. So it your case it would be set to `http://www.data.qld.gov.au/data-sets/water/gregory-environmental.csv?timestamp=2025-01-16T10:15:01+10:00` as I think your `ckan.site_url` is set to `http://www.data.qld.gov.au`

duttonw: Hi @kowh-ai ,

http://www.data.qld.gov.au/data-sets/water/gregory-environmental.csv?timestamp=2025-01-16T10:15:01+10:00 is a 404, since its not on www.data.qld.gov.au, but on apps.des.qld.gov.au (non ckan endpoint)

With the new function ``def get_ckan_url():``
just update it to take a url. And it would alter it if incoming url 'matches' ``ckan.site_url`` at the start of the string and replace it with ``ckanext.xloader.site_url`` if not empty/blank string. (or if no domain is set, inject). 

This will still need at least another 5 unit tests for this new function to 'verify' that it does what we expect it to do. i.e. not break external data assets. 

---

somethings i've been thinking for a while is that if you have the 'archiver' plugin installed, like how there is a use case to 'wait' for the validation plugin to complete successfully. If the archiver has 'saved' the resource locally, then we could/should use that ingest into the datastore. The only negative is like these external csv files which are updated adhoc and the 'manual datastore upload' won't collect the latest like it does now. (harvester normally is set up to run on initial upload and then can be setup to run adhoc { for our instance monthly } , this is too slow and will cause frustration for authors who wish to have the latest from an external csv )

ThrawnCA: This whole section should probably be moved down ten lines or so, to the block where `url_type` is `upload`. When the URL type is `link` the site URL is not involved in downloading it.

kowh-ai: @duttonw - What about the other callbacks to CKAN ie: when these 2 functions call `callback_xloader_hook(result_url, api_key, job_dict)`

1. `xloader_data_into_datastore()`
2. `xloader_data_into_datastore_()`

The `result_url` parameter will need to change in a docker network to access the internal CKAN container ie: http://ckan-dev:5000/<etc>

In your case _could_ the `ckan.site_url` be different to where you want to access the CKAN URL from within the XLoader code? like when you are downloading the resource file? In my case I need to set the CKAN URL to `ckanext.xloader.site_url`

ckanext/xloader/action.py
```diff
@@ -140,17 +140,18 @@ def xloader_submit(context, data_dict):
...
-            'task_created': task['last_updated'],
-            'original_url': resource_dict.get('url'),
-        }
+        "api_key": utils.get_xloader_user_apitoken(),
+        "job_type": "xloader_to_datastore",
+        "result_url": callback_url,
+        "metadata": {
+            "ignore_hash": data_dict.get("ignore_hash", False),
+            "ckan_url": config.get("ckanext.xloader.site_url")
+            or config["ckan.site_url"],
```
ThrawnCA, line 149: I think this should be indented or placed on the previous line.

ckanext/xloader/utils.py
```diff
@@ -107,6 +110,61 @@ def get_xloader_user_apitoken():
...
+    
+    Args:
+        result_url (str): The original URL to potentially modify
+        ckan_url (str): The base CKAN site URL to compare against
+    Returns:
+        str: The modified URL if base URLs differ, otherwise returns original URL unchanged 
+    """
+    parsed_url = urlparse(result_url)
+    base_url = f"{parsed_url.scheme}://{parsed_url.netloc}"
+    path_url = parsed_url.path
```
ThrawnCA, line 128: `path_url` is only needed if `base_url != ckan_url`, no point assigning it here.

ckanext/xloader/utils.py
```diff
@@ -107,6 +110,61 @@ def get_xloader_user_apitoken():
...
+    Args:
+        orig_ckan_url (str): The original CKAN URL to potentially modify
+    Returns:
+        str: Either the modified URL with new base URL from xloader_site_url, 
+             or the original URL if conditions aren't met
+    """
+    xloader_site_url = config.get('ckanext.xloader.site_url')
+    ckan_site_url = config.get('ckan.site_url')
+
+    parsed_url = urlparse(orig_ckan_url)
```
ThrawnCA, line 153: This isn't needed until we're inside the conditional.

kowh-ai: Thanks @ThrawnCA  - Whereabouts? AFAICT I'm seeing `xloader_site_url`,  `ckan_site_url` and `parsed_url` needed for the first conditional ie: `if base_url == ckan_site_url and xloader_site_url:` (`parsed_url` is needed to determine `base_url`)

ThrawnCA: Ah, I missed the usage inside the f-string.

kowh-ai:
@duttonw and @ThrawnCA - Could you please take a fresh look at the changes to this PR I've implemented and either approve or suggest more updates - thank you both!

ckanext/xloader/jobs.py
```diff
@@ -276,9 +280,11 @@ def _download_resource_data(resource, data, api_key, logger):
     data['datastore_contains_all_records_of_source_file'] = False
     which will be saved to the resource later on.
     '''
-    # check scheme
+ 
     url = resource.get('url')
-    url_parts = urlsplit(url)
+    url = modify_resource_url(url)
```
ThrawnCA, line 285: I'm still not happy about this being applied to every resource download. How is this meant to handle cases where we _want_ to download from an external site?

ckanext/xloader/utils.py
```diff
@@ -107,6 +110,60 @@ def get_xloader_user_apitoken():
...
+    2. An xloader_site_url is configured in the settings
+    
+    Args:
+        orig_ckan_url (str): The original CKAN URL to potentially modify
+    Returns:
+        str: Either the modified URL with new base URL from xloader_site_url, 
+             or the original URL if conditions aren't met
+    """
+    xloader_site_url = config.get('ckanext.xloader.site_url')
+    ckan_site_url = config.get('ckan.site_url')
```
ThrawnCA, line 150: Maybe check that the two values are not equal before proceeding.

Note that the validator snippet in `config.yaml` will cause `ckanext.xloader.site_url` to be equal to `ckan.site_url` if not supplied.

ckanext/xloader/utils.py
```diff
@@ -107,6 +110,60 @@ def get_xloader_user_apitoken():
...
+
+    parsed_url = urlparse(orig_ckan_url)
+    base_url = f"{parsed_url.scheme}://{parsed_url.netloc}"
+
+    # If the base URL matches the CKAN site URL and xloader_site_url is set, modify the URL
+    if base_url == ckan_site_url and xloader_site_url:
```
ThrawnCA, line 156: `xloader_site_url` will pretty much always be truthy, because its default value is `ckan.site_url`. We should instead be checking if they are equal.

ThrawnCA (changes requested):

duttonw:
Hi @kowh-ai , Sorry things have been busy, I have started to add more tests to test edge cases for this code per https://github.com/pwalsh/ckanext-xloader/pull/1 but I was re-tasked on other things so did not have time to 'fix' the code to pass said tests (or double verify my tests are 'valid'). 

I'll try and put a couple of hours into this since this is a very critical feature we all want and we need to ensure its done right.

duttonw:
Hi @kowh-ai ,

I've updated my tests and refactored the utils function, then had to rebase your branch as it was out of sync with master.

If you are happy, we can close @234  as its superseded by  #243
https://github.com/ckan/ckanext-xloader/pull/243

kowh-ai:
Thanks!!! Will pick this back up Wednesday as I'm BER-AKL at the mo

duttonw:
Hi @kowh-ai , 
I'm holding off merging https://github.com/ckan/ckanext-xloader/pull/243 until you get time to review it.

regards,

@duttonw

kowh-ai:
@duttonw - arrived in New Zealand , will review and fully test this week! thanks very much for your work here!

duttonw:
Closing, merged via rebased branch #243 thanks for the contribution @pwalsh and @kowh-ai

PR #237, opened 2024-12-09
Validation Extension Support (Squashed)

JVickery-TBS:
squash(pr): 200;

- Squashed version of https://github.com/ckan/ckanext-xloader/pull/200.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ Compatibility with core CKAN versions:
   | CKAN version   | Compatibility                                                                                                  |
   | -------------- | -------------------------------------------------------------------------------------------------------------- |
   | 2.7            | no longer supported (last supported version: 0.12.2)                                                           |
-  | 2.8            | no longer supported (last supported version: 0.12.2)                                                           | 
+  | 2.8            | no longer supported (last supported version: 0.12.2)                                                           |
   | 2.9            | yes (Python3) (last supported version for Python 2.7: 0.12.2)), Must: `pip install "setuptools>=44.1.0,<71"`   |
   | 2.10           | yes                                                                                                            |
   | 2.11           | yes                                                                                                            |
@@ -189,7 +189,7 @@ expect European (day-first) dates, you could add to `postgresql.conf`:
 
     datestyle=ISO,DMY
 
-All configurations below are defined in the 
+All configurations below are defined in the
 [config_declaration.yaml](ckanext/xloader/config_declaration.yaml) file.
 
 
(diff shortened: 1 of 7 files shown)


ckanext/xloader/action.py
```diff
@@ -166,15 +173,20 @@ def xloader_submit(context, data_dict):
         job = enqueue_job(
             jobs.xloader_data_into_datastore, [data], queue=custom_queue,
             title="xloader_submit: package: {} resource: {}".format(resource_dict.get('package_id'), res_id),
-            rq_kwargs=dict(timeout=timeout)
+            rq_kwargs=dict(timeout=timeout, at_front=sync)
```
duttonw, line 176: what version had at_front added to the api?

ckanext/xloader/utils.py
```diff
@@ -2,6 +2,7 @@
 
 import json
 import datetime
+from rq import get_current_job
```
duttonw, line 5: you added get_current_job, but i don't see any reference in this file, is it meant to be part of anther enhancement?

ckanext/xloader/action.py
JVickery-TBS: @duttonw the `rq_kwargs` are to pass to the `rq` lib. So not sure when rq added at_front

ckanext/xloader/utils.py
JVickery-TBS: @duttonw my bad, it was for a really bad implementation of direct chaining with the validation job and xloader job. Before I found the `at_front` from the RQ lib. import has been removed.

ckanext/xloader/action.py
JVickery-TBS: @duttonw https://github.com/rq/rq/blob/master/CHANGES.md#050 <- okay so it was added in 2015. In version `0.5.0` and CKAN 2.8 was using `rq==0.6.0` so we should be all good with this usage.

README.md
```diff
@@ -409,6 +409,34 @@ Controls whether or not the status badges display all of the statuses. By defaul
...
+
+Example:
+
+```
+ckanext.xloader.validation.enforce_schema = False
+```
+
+Default value: `True`
+
+Controls whether or not a resource requires a Validation Schema to be present from the ckanext-validation plugin to be XLoadered.
```
ThrawnCA, line 438: It would be good to document how this interacts with `requires_successful_report`. What happens when one is True and the other is False?

PR #241, opened 2025-01-27
Avoid loading R/W datastore resources

wardi:
Fix for compatibility with CKAN Table Designer feature

### DIFF
--- a/ckanext/xloader/jobs.py
+++ b/ckanext/xloader/jobs.py
@@ -19,7 +19,7 @@ from rq.timeouts import JobTimeoutException
 import sqlalchemy as sa
 
 from ckan import model
-from ckan.plugins.toolkit import get_action, asbool, enqueue_job, ObjectNotFound, config
+from ckan.plugins.toolkit import get_action, asbool, enqueue_job, ObjectNotFound, config, h
 
 from . import db, loader
 from .job_exceptions import JobError, HTTPError, DataTooBigError, FileCouldNotBeLoadedError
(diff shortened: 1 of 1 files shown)


ThrawnCA (approved):

ckanext/xloader/jobs.py
```diff
@@ -178,8 +178,8 @@ def xloader_data_into_datastore_(input, job_dict, logger):
     logger.info('Express Load starting: %s', resource_ckan_url)
 
     # check if the resource url_type is a datastore
-    if resource.get('url_type') == 'datastore':
-        logger.info('Ignoring resource - url_type=datastore - dump files are '
+    if resource.get('url_type') in h.datastore_rw_resource_url_types():
```
duttonw, line 181: Needs backwards support for for 2.10.x since datastore_rw_resource_url_types only exists in 2.11.

https://github.com/ckan/ckan/blob/ckan-2.11.1/ckanext/datastore/helpers.py#L242 (exists since May 2023)
https://github.com/ckan/ckan/blob/ckan-2.10.6/ckanext/datastore/helpers.py (does not include 'datastore_rw_resource_url_types'  )

i.e.
```
if hasattr(h, "datastore_rw_resource_url_types"):
             datastore_rw_resource_url_types = h.datastore_rw_resource_url_types()
else:
             #fallback for 2.10.x or older.
             datastore_rw_resource_url_types = ['datastore']
           
if resource.get('url_type') in datastore_rw_resource_url_types :
```

duttonw (changes requested):
Testing is failing for 2.10.x

Provided comment to make it work with both.

ckanext/xloader/jobs.py
wardi, line 181: just reformatting @duttonw 's code as a suggestion
```suggestion
    if hasattr(h, "datastore_rw_resource_url_types"):
        datastore_rw_resource_url_types = h.datastore_rw_resource_url_types()
    else:
        #fallback for 2.10.x or older.
        datastore_rw_resource_url_types = ['datastore']
           
    if resource.get('url_type') in datastore_rw_resource_url_types:
```

duttonw (approved):

PR #243, opened 2025-03-03
Feature/custom xloader site url rebased

duttonw:
supersedes: pwalsh:feature/custom-xloader-site-url #234

We want to be able to communicate within a docker network without using the public site_url. This is a minimal proof of concept for achieving this.

Update: There are 2 scenario's where XLoader communicates to CKAN

Scenario 1: When it downloads the resource file from CKAN ie: in _download_resource_data()
Scenario 2: When it updates the status of the XLoader job ie: in callback_xloader_hook()

If the config option ckanext.xloader.site_url is set then all communication from XLoader to CKAN will use this URL. In scenario 1, if it's not set then the ckan.site_url config option will be used unless the download ie: _download_resource_data() original_url is different to ckan_url (ie: ckan.site_url) and in that case the _download_resource_data() download will use the original_url but all other communication back to CKAN (ie: scenario 2) will use the ckan.site_url option

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -34,6 +34,7 @@ jobs:
 
   validateVersion:
     runs-on: ubuntu-latest
+    if: github.repository == 'ckan/ckanext-xloader'
     steps:
       - uses: actions/checkout@v4
 
@@ -52,79 +53,10 @@ jobs:
             exit 1
           fi
 
-  lint:
-    needs: validateVersion
-    if: github.repository == 'ckan/ckanext-xloader'
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v4
-
-      - uses: actions/setup-python@v5
-        with:
-          python-version: '3.10'
-
-      - name: Install requirements
-        run: pip install flake8 pycodestyle
-
-      - name: Check syntax
-        run: flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics --extend-exclude ckan
-
   test:
-    needs: lint
-    strategy:
-      matrix:
-        include: #ckan-image see https://github.com/ckan/ckan-docker-base, ckan-version controls other image tags
-          - ckan-version: "2.11"
-            ckan-image: "2.11-py3.10"
-          - ckan-version: "2.10"
-            ckan-image: "2.10-py3.10"
-          #- ckan-version: "master" Publish does not care about master
-          #  ckan-image: "master"
-      fail-fast: false
-
-    name: CKAN ${{ matrix.ckan-version }}
-    runs-on: ubuntu-latest
-    container:
-      image: ckan/ckan-dev:${{ matrix.ckan-image }}
-      options: --user root
-    services:
-      solr:
-        image: ckan/ckan-solr:${{ matrix.ckan-version }}-solr9
-      postgres:
-        image: ckan/ckan-postgres-dev:${{ matrix.ckan-version }}
-        env:
-          POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: postgres
-          POSTGRES_DB: postgres
-        ports:
-          - 5432:5432
-        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
-      redis:
-          image: redis:3
-    env:
-      CKAN_SQLALCHEMY_URL: postgresql://ckan_default:pass@postgres/ckan_test
-      CKAN_DATASTORE_WRITE_URL: postgresql://datastore_write:pass@postgres/datastore_test
-      CKAN_DATASTORE_READ_URL: postgresql://datastore_read:pass@postgres/datastore_test
-      CKAN_SOLR_URL: http://solr:8983/solr/ckan
-      CKAN_REDIS_URL: redis://redis:6379/1
-
-    steps:
-    - uses: actions/checkout@v4
-    - if: ${{ matrix.ckan-version == 2.9 }}
-      run: pip install "setuptools>=44.1.0,<71"
-    - name: Install requirements
-      run: |
-        pip install -r requirements.txt
-        pip install -r dev-requirements.txt
-        pip install -e .
-        pip install -U requests[security]
-        # Replace default path to CKAN core config file with the one on the container
-        sed -i -e 's/use = config:.*/use = config:\/srv\/app\/src\/ckan\/test-core.ini/' test.ini
-    - name: Setup extension (CKAN >= 2.9)
-      run: |
-        ckan -c test.ini db init
-    - name: Run tests
-      run: pytest --ckan-ini=test.ini --cov=ckanext.xloader --disable-warnings ckanext/xloader/tests
+    needs: validateVersion
+    name: Test
+    uses: ./.github/workflows/test.yml # Call the reusable workflow
 
   publishSkipped:
     if: github.repository != 'ckan/ckanext-xloader'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,7 +35,7 @@ jobs:
             experimental: true  # master is unstable, good to know if we are compatible or not
       fail-fast: false
 
-    name: CKAN ${{ matrix.ckan-version }}
+    name: ${{ matrix.experimental && '**Fail_Ignored** ' || '' }} CKAN ${{ matrix.ckan-version }}
     runs-on: ubuntu-latest
     container:
       image: ckan/ckan-dev:${{ matrix.ckan-image }}
@@ -65,12 +65,7 @@ jobs:
     - uses: actions/checkout@v4
       continue-on-error: ${{ matrix.experimental }}
 
-    - name: Pin setuptools for ckan 2.9 only
-      if: ${{ matrix.ckan-version == 2.9 }}
-      run: pip install "setuptools>=44.1.0,<71"
-      continue-on-error: ${{ matrix.experimental }}
-
-    - name: Install requirements
+    - name: ${{ matrix.experimental && '**Fail_Ignored** ' || '' }} Install requirements
       continue-on-error: ${{ matrix.experimental }}
       run: |
         pip install -r requirements.txt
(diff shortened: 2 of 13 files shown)


ckanext/xloader/plugin.py
```diff
@@ -76,14 +76,6 @@ def configure(self, config_):
         else:
             self.ignore_hash = False
 
-        for config_option in ("ckan.site_url",):
```
duttonw, line 79: ckan.site_url is mandatory in 2.10+ and defaults to 127.0.0.1:5000 if not set, we don't need to config validate this.

ckanext/xloader/config_declaration.yaml
```diff
@@ -2,6 +2,23 @@ version: 1
...
   - annotation: ckanext-xloader settings
     options:
+      - key: ckanext.xloader.site_url
+        example: http://ckan-dev:5000
+        default:
+        description: |
+            Provide an alternate site URL for the xloader_submit action.
+            This is useful, for example, when the site is running within a docker network.
+            Note: This setting will not alter path. i.e ckan.root_path
+        validators: configured_default("ckan.site_url",None)
```
ThrawnCA, line 12: IMO it's better to just leave it blank by default. It's simpler for code to say "fall back to site_url if the XLoader URL is falsy".

ckanext/xloader/tests/test_jobs.py
```diff
@@ -71,17 +71,17 @@ def data(create_with_upload, apikey):
...
-            'set_url_type': False,
-            'task_created': datetime.utcnow().isoformat(),
-            'original_url': resource["url"],
-        }
+        "api_key": apikey,
+        "job_type": "xloader_to_datastore",
+        "result_url": callback_url,
+        "metadata": {
+            "ignore_hash": True,
+            "ckan_url": toolkit.config.get("ckanext.xloader.site_url", toolkit.config.get("ckan.site_url")),
```
ThrawnCA: This fallback only makes sense if we don't provide a default for `ckanext.xloader.site_url`.

ckanext/xloader/tests/test_jobs.py
```diff
@@ -103,6 +103,66 @@ def test_xloader_data_into_datastore(self, cli, data):
         resource = helpers.call_action("resource_show", id=data["metadata"]["resource_id"])
         assert resource["datastore_contains_all_records_of_source_file"]
 
+    def test_download_resource_data_with_ckanext_xloader_site_url(self, cli, data):
+        # Set the ckanext.xloader.site_url in the config
+        with mock.patch.dict(toolkit.config, {'ckanext.xloader.site_url': 'http://xloader-site-url'}):
```
ThrawnCA, line 108: This can be done more cleanly via `@pytest.mark.ckan_config` test annotations.

ckanext/xloader/tests/test_utils.py
```diff
@@ -0,0 +1,62 @@
+import pytest
+from unittest.mock import patch
+from ckan.plugins import toolkit
+from ckanext.xloader import utils
+
+
+@pytest.mark.parametrize("result_url, ckan_url, expected", [
```
ThrawnCA, line 7: What about empty/blank/`None` inputs?

ckanext/xloader/utils.py
```diff
@@ -175,6 +178,68 @@ def get_xloader_user_apitoken():
...
+
+    Args:
+        input_url (str): The original CKAN URL to potentially modify
+    Returns:
+        str: Either the modified URL with new base URL from xloader_site_url,
+             or the original URL if conditions aren't met
+    """
+
+    xloader_site_url = config.get('ckanext.xloader.site_url')
+    if not xloader_site_url:
```
ThrawnCA, line 223: This only makes sense if we don't configure a default value.

ThrawnCA (changes requested):

ckanext/xloader/config_declaration.yaml
duttonw: May just remove the validators, its not required. (unless you know of one that is testing for url optionally)

ckanext/xloader/tests/test_jobs.py
duttonw: removed

ckanext/xloader/tests/test_utils.py
duttonw: will add some more.

ckanext/xloader/config_declaration.yaml
duttonw: done

ckanext/xloader/tests/test_jobs.py
duttonw: done

ckanext/xloader/tests/test_jobs.py
duttonw: I've not touched @kowh-ai 's tests. @ThrawnCA should I improve them?

ckanext/xloader/tests/test_utils.py
duttonw: done.

ckanext/xloader/utils.py
duttonw: default value is removed, so its None. 
Exit early no point wasting cpu cycles on the rest of the function.

ckanext/xloader/tests/test_jobs.py
ThrawnCA: I think it's better to use the annotations than `with`.

I'm not really reviewing this in conjunction with the other PR, I'm just looking at it on its own merits.

ckanext/xloader/tests/test_utils.py
```diff
@@ -0,0 +1,70 @@
...
+                     "ckanext.xloader.site_url_ignore_path_regex": "(/blackListedPathToS3HostOrigin|/anotherpath)"}):
+        response = utils.modify_input_url(input_url)
+        if is_altered:
+            assert response == expected
+        else:
+            assert response == input_url
+
+
+
+def test_modify_input_url_no_xloader_site():
```
ThrawnCA, line 67: Why is this a separate test instead of just another set of parameterised inputs?

duttonw: it's also in the parameterised, but it may still be nice to keep seperate for 'clarity'

ThrawnCA (approved):

.github/workflows/publish.yml
```diff
@@ -52,79 +53,10 @@ jobs:
...
-        # Replace default path to CKAN core config file with the one on the container
-        sed -i -e 's/use = config:.*/use = config:\/srv\/app\/src\/ckan\/test-core.ini/' test.ini
-    - name: Setup extension (CKAN >= 2.9)
-      run: |
-        ckan -c test.ini db init
-    - name: Run tests
-      run: pytest --ckan-ini=test.ini --cov=ckanext.xloader --disable-warnings ckanext/xloader/tests
+    needs: validateVersion
+    name: Test
+    uses: ./.github/workflows/test.yml # Call the reusable workflow
```
duttonw, line 59: Use test.yml so we don't need to duplicate how we test again.

.github/workflows/test.yml
```diff
@@ -35,7 +35,7 @@ jobs:
             experimental: true  # master is unstable, good to know if we are compatible or not
       fail-fast: false
 
-    name: CKAN ${{ matrix.ckan-version }}
+    name: ${{ matrix.experimental && '**Fail_Ignored** ' || '' }} CKAN ${{ matrix.ckan-version }}
```
duttonw, line 38: duplicate from another repo to give info that its 'ignoring' errors.

.github/workflows/test.yml
```diff
@@ -80,16 +75,20 @@ jobs:
...
         sed -i -e 's/use = config:.*/use = config:\/srv\/app\/src\/ckan\/test-core.ini/' test.ini
 
-    - name: Setup extension
+    - name: ${{ matrix.experimental && '**Fail_Ignored** ' || '' }} Setup extension
       continue-on-error: ${{ matrix.experimental }}
       run: |
         ckan -c test.ini db init
+        ckan -c test.ini user add ckan_admin email=ckan_admin@localhost password="AbCdEf12345!@#%"
+        ckan -c test.ini sysadmin add ckan_admin
+        ckan config-tool test.ini "ckanext.xloader.api_token=$(ckan -c test.ini user token add ckan_admin xloader | tail -n 1 | tr -d '\t')"
```
duttonw, line 84: configure proper xloader api token for test based on what would be going into the docker container.

README.md
```diff
@@ -151,6 +151,7 @@ To install XLoader:
     execute jobs against the server:
 
         ckanext.xloader.api_token = <your-CKAN-generated-API-Token>
+        ckan config-tool test.ini "ckanext.xloader.api_token=$(ckan -c test.ini user token add ckan_admin xloader | tail -n 1 | tr -d '\t')"
```
duttonw, line 154: todo: readme update on new ckan config options for custom xloader site url

ckanext/xloader/config_declaration.yaml
```diff
@@ -14,9 +30,9 @@ groups:
         example: eyJ0eXAiOiJKV1QiLCJh.eyJqdGkiOiJ0M2VNUFlQWFg0VU.8QgV8em4RA
         description: |
             Uses a specific API token for the xloader_submit action instead of the
-            apikey of the site_user. Will be mandatory when dropping support for
-            CKAN 2.9.
-        required: false
+            apikey of the site_user.
+        default: 'NOT_SET'
+        required: true
```
duttonw, line 35: dropped 2.9, set to mandatory and give it a default we can throw validation errors on. (i.e. stop the chicken and egg problem that you need a running ckan to create the api key to reference in the config)

ckanext/xloader/tests/test_action.py
```diff
@@ -117,10 +118,24 @@ def test_status(self):
...
 
-    def test_xloader_user_api_token_defaults_to_site_user_apikey(self):
-        api_token = get_xloader_user_apitoken()
-        site_user = helpers.call_action("get_site_user")
-        assert api_token == site_user["apikey"]
+
+    def test_xloader_user_api_token_from_config(self):
+        sysadmin = factories.SysadminWithToken()
+        apikey = sysadmin["token"]
+        with mock.patch.dict(toolkit.config, {'ckanext.xloader.api_token': apikey}):
```
duttonw, line 125: this is to mimic what would be in the config.  I ran out of time to work out how to look up a key inside pytest or hope that the db setup at the start was not 'cleansed'

ckanext/xloader/utils.py
```diff
@@ -167,12 +170,73 @@ def get_xloader_user_apitoken():
     method returns the api_token set in the config file and defaults to the
     site_user.
     """
-    api_token = p.toolkit.config.get('ckanext.xloader.api_token', None)
-    if api_token:
+    api_token = p.toolkit.config.get('ckanext.xloader.api_token')
+    if api_token and api_token != 'NOT_SET':
         return api_token
+    raise p.toolkit.ValidationError({u'ckanext.xloader.api_token': u'NOT_SET, please provide valid api token'})
```
duttonw, line 176: hope this change catches miss configurations where xloader does it upload into the datastore but never 'completes' the job.

ckanext/xloader/tests/test_action.py
ThrawnCA: For dynamic token values this approach does make sense.

ThrawnCA (approved):

kowh-ai (approved):
@duttonw - tested fine for a CKAN  (2.11 and 2.10) Docker environment that replaces DataPusher with this XLoader branch. Thank You!

PR #245, opened 2025-03-03
Fix: Support CKAN 2.11.x Config apitoken_header_name - default to 'Authorization'

Engerrs:
This PR adds support for the new "apitoken_header_name" CKAN Config, in which you can specify alternative header for CKAN API Tokens.
This is an issue, when your CKAN instances are under Basic HTTP Authentication, which breaks the default "Authorization" header and as a result, uploading doesn't work and return 403 HTTP.
Those changes make sure that Xloader takes into account the changed Header for accepting CKAN API Keys.


duttonw:
Hi @Engerrs ,

Thanks for the contribution.

Regards,

@duttonw

duttonw (approved):

PR #252, opened 2025-05-29
fix: add pagination to 'submit all' to process all datasets

dsanchezmatilla:
This PR adds a loop to paginate through all datasets in batches of 1000 using package_search, so that all datasets are submitted to the xloader, not just the first 1000.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -54,7 +54,7 @@ jobs:
           - 5432:5432
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
       redis:
-          image: redis:3
+          image: redis:7
     env:
       CKAN_SQLALCHEMY_URL: postgresql://ckan_default:pass@postgres/ckan_test
       CKAN_DATASTORE_WRITE_URL: postgresql://datastore_write:pass@postgres/datastore_test


ckanext/xloader/command.py
```diff
@@ -46,14 +46,27 @@ def _submit_all(self, sync=False, queue=None):
...
+            paged_response = tk.get_action('package_search')({'ignore_auth': True}, arguments)
+            package_list.extend([pkg['id'] for pkg in paged_response['results']])
+            arguments['start'] += 1000
+
         print('Processing %d datasets' % len(package_list))
-        user = tk.get_action('get_site_user')(
-            {'ignore_auth': True}, {})
-        for p_id in package_list:
-            self._submit_package(p_id, user, indent=2, sync=sync, queue=queue)
+        check_start = input('This action could take a few minuts depending on the number of DataSets:\nDid you want to start the process? y/N\n')
```
duttonw, line 62: minuts/minutes

ckanext/xloader/command.py
```diff
@@ -46,14 +46,27 @@ def _submit_all(self, sync=False, queue=None):
...
+
+        if check_start == 'y':
+            user = tk.get_action('get_site_user')({'ignore_auth': True}, {})
+            for p_id in package_list:
+                self._submit_package(p_id, user, indent=2, sync=sync, queue=queue)
+        else:
+            print('Submit all process stoped')
```
duttonw, line 69: stoped/stopped

ckanext/xloader/command.py
```diff
@@ -46,14 +46,27 @@ def _submit_all(self, sync=False, queue=None):
...
+        arguments = {'include_private': True, 'start': 0, 'rows': 10}
+
+        response = tk.get_action('package_search')({'ignore_auth': True}, {'include_private': True})
+        num_pages = (response['count'] + 999) // 1000
+        arguments['rows'] = 1000
+        package_list = []
+
+        for page in range(0, num_pages):
+            paged_response = tk.get_action('package_search')({'ignore_auth': True}, arguments)
+            package_list.extend([pkg['id'] for pkg in paged_response['results']])
```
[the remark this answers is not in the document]
duttonw: on @ThrawnCA input, you could get a total number of packages in the system (without all the data), then ask the question. 

If yes, then do the pagination and submit the jobs so you can discard/release memory. This becomes very required when you have over 100,000+ datasets and working on small containers.

duttonw (changes requested):
Its a good contribution but does require spelling corrections and a slight change in logic.

Overall, a good PR :)

dsanchezmatilla:
@duttonw and @ThrawnCA , thanks for the advice about possible memory management issues with my code when working with large numbers of datasets. I'm currently working on a solid fix for the problem.

PR #256, opened 2025-07-31
perf: implement chunked fulltext indexing for large datasets

### LINKED ISSUE #255: Implement chunked processing for fulltext indexing to prevent timeouts on large datasets
**Description:**

The _populate_fulltext function currently executes a single UPDATE query to populate the _full_text column for all rows at once. This causes database timeouts and performance issues when processing very large datasets (4GB+ files with millions of rows).

**Problem:**

Single UPDATE query on entire table causes statement timeouts

Large datasets can exhaust database memory during tsvector generation

Long-running queries create lock contention blocking other operations

Transaction logs can overflow on massive updates

No progress visibility during long indexing operations

**Current Behavior:**

```
UPDATE table SET _full_text = to_tsvector(concatenated_columns);
-- Processes ALL rows in one operation
```

**Proposed Solution:**
Partition the fulltext indexing into configurable chunks using _id ranges:
```

UPDATE table SET _full_text = to_tsvector(concatenated_columns) 
WHERE _id BETWEEN 1 AND 100000;

UPDATE table SET _full_text = to_tsvector(concatenated_columns) 
WHERE _id BETWEEN 100001 AND 200000;
-- Continue until all rows processed
```

**Implementation Details:**

Add configurable chunk size via ckanext.xloader.search_update_chunks (default: 100,000)

Process rows in batches based on _id ranges

Add progress logging for each chunk

Continue processing remaining chunks if individual chunks fail

Maintain existing functionality for smaller datasets

**Benefits:**

Prevents database timeouts on large files

Reduces memory pressure during indexing

Provides progress visibility through logging

Allows resumable processing if interrupted

Maintains performance for smaller datasets

**Configuration:**

# Optional: Configure chunk size (default: 100000)
ckanext.xloader.search_update_chunks = 50000

**Files to Modify:**

ckanext/xloader/loader.py - Update _populate_fulltext function

**Testing:**

Test with small datasets (< 1000 rows) - should work as before

Test with large datasets (> 1M rows) - should complete without timeout

Test chunk size configuration

Verify progress logging works correctly

cgoldshtein:
## Description
Implement chunked processing for fulltext indexing to prevent timeouts on large datasets.

Fixes #255 

## Changes
- Replace single UPDATE query with configurable chunked processing
- Add progress logging for each processed chunk
- Add new config option: ckanext.xloader.search_update_chunks

### DIFF
--- a/ckanext/xloader/loader.py
+++ b/ckanext/xloader/loader.py
@@ -659,30 +659,86 @@ def _enable_fulltext_trigger(connection, resource_id):
         .format(table=identifier(resource_id, True))))
 
 
-def _populate_fulltext(connection, resource_id, fields):
-    '''Populates the _full_text column. i.e. the same as datastore_run_triggers
-    but it runs in 1/9 of the time.
+def _get_rows_count_of_resource(connection, table):
+    count_query = ''' SELECT count(_id) from {table} '''.format(table=table)
+    results = connection.execute(count_query)
+    rows_count = int(results.first()[0])
+    return rows_count
 
-    The downside is that it reimplements the code that calculates the text to
-    index, breaking DRY. And its annoying to pass in the column names.
 
-    fields: list of dicts giving the each column's 'id' (name) and 'type'
-            (text/numeric/timestamp)
+def _populate_fulltext(connection, resource_id, fields, logger):
+    '''Populates the _full_text column for full-text search functionality.
+    
+    This function creates a PostgreSQL tsvector (text search vector) for each row
+    by concatenating all non-system columns. It's equivalent to datastore_run_triggers
+    but runs approximately 9x faster by using direct SQL updates.
+    
+    To handle very large datasets (e.g., 4GB+ files with millions of rows), the update
+    operation is partitioned into chunks to prevent:
+    - Database statement timeouts
+    - Memory exhaustion
+    - Lock contention that could block other operations
+    - Transaction log overflow
+    
+    The chunking mechanism processes rows in batches based on their _id values,
+    with chunk size configurable via 'ckanext.xloader.search_update_chunks' 
+    (default: 100,000 rows per chunk).
+    
+    Args:
+        connection: Database connection object
+        resource_id (str): The datastore table identifier
+        fields (list): List of dicts with column 'id' (name) and 'type' 
+                      (text/numeric/timestamp)
+        logger: Logger instance for progress tracking
+        
+    Note:
+        This reimplements CKAN's text indexing logic for performance,
+        breaking DRY principle but providing significant speed improvements.
     '''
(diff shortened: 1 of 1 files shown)


ckanext/xloader/loader.py
```diff
@@ -659,30 +659,86 @@ def _enable_fulltext_trigger(connection, resource_id):
...
-    fields: list of dicts giving the each column's 'id' (name) and 'type'
-            (text/numeric/timestamp)
+def _get_rows_count_of_resource(connection, table):
+    count_query = ''' SELECT count(_id) from {table} '''.format(table=table)
+    results = connection.execute(count_query)
+    rows_count = int(results.first()[0])
+    return rows_count
+
+
+def _populate_fulltext(connection, resource_id, fields, logger):
```
ThrawnCA, line 669: `logger` is not provided by the calling code.

ckanext/xloader/loader.py
```diff
@@ -659,30 +659,86 @@ def _enable_fulltext_trigger(connection, resource_id):
...
+        rows_count = _get_rows_count_of_resource(connection, identifier(resource_id))
+    except Exception as e:
+        rows_count = ''
+        logger.info("Failed to get resource rows count: {} ".format(str(e)))
+        raise
+        
+    if rows_count:
+        # Configure chunk size - prevents timeouts and memory issues on large datasets
+        # Default 100,000 rows per chunk balances performance vs. resource usage
+        chunks = int(config.get('ckanext.xloader.search_update_chunks', 100000))
```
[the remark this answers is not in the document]
cgoldshtein: Done.

ckanext/xloader/loader.py
cgoldshtein: Fixed.

wardi:
How large of a table can now be loaded with this change?

duttonw:
Thats a very good question and I think that comes down to how beefy the database is as well as the worker now. Chunking the data ensures you don't hit ram/contention states.

wardi:
@duttonw I'm really happy to see this change. We've got some 11M row datasets that wouldn't load with xloader likely due to this issue.

So far we've worked around the issue by using table designer for the resources (to use the most efficient column types) and loading the data with the API but it would be simpler for our client to use xloader since the schema changes fairly often and we typically can't make use of the partial update ability.

The second problem has been slow pagination in streaming downloads of such a large dataset, this is being addressed in https://github.com/ckan/ckan/pull/9028

The third problem is not having a way to define composite indexes with the datastore API. Sorting efficiently requires an index with the column(s) being sorted followed by the `_id` field because that's used to provide stable ordering for pagination. We've worked around this by manually adding indexes to these large tables with `psql`, but hope to work on extending the datastore_create API for adding indexes soon.

PR #275, opened 2026-04-15
Implement a new Interface function for datastore_before_update

avdata99:
Adding the `datastore_before_update` hook to `IXloader`
I need to hook into the datastore updates to detect field changes
                                                                                
This PR adds a new optional method to the IXloader interface so plugins can observe schema changes before XLoader modifies the DataStore table.
                                                                                                                                                                                      
The `datastore_before_update` call includes:
  - existing_fields: list of current DataStore field dicts (each has id, type, and optionally info), or None if the table does not yet exist. The internal `_id` column is stripped.
  - new_headers: list of field dicts about to be written.

 - [x] tests added. Ran [here](https://github.com/avdata99/ckanext-xloader/actions/runs/24473976023) because of this repo permission

### DIFF
--- a/ckanext/xloader/interfaces.py
+++ b/ckanext/xloader/interfaces.py
@@ -47,3 +47,30 @@ class IXloader(Interface):
             the resource that was uploaded
         """
         pass
+
+    def datastore_before_update(self, resource_id, existing_info, new_headers):
+        """ Called by the loader just before it is about to modify the
+        DataStore table for a resource (truncate, drop+recreate, or create).
+        It allows plugins to inspect the difference between the current
+        DataStore columns and the ones detected in the incoming file, for
+        example to log an activity when columns are added, removed or
+        renamed.
+
+        :param resource_id: the ID of the resource whose DataStore table is
+            about to be updated.
+        :type resource_id: string
+
+        :param existing_info: a mapping of ``{field_id: info_dict}`` built
+            from the existing DataStore table's Data Dictionary, or ``None``
+            if the DataStore table does not yet exist.
+        :type existing_info: dict or None
+
+        :param new_headers: the list of field dicts that will be written to
+            the DataStore. Each dict has at least an ``id`` and ``type``
+            key, and may include an ``info`` dict for fields that already
+            existed.
+        :type new_headers: list of dicts
+
+        The return value is ignored.
+        """
+        pass
--- a/ckanext/xloader/interfaces.py
+++ b/ckanext/xloader/interfaces.py
@@ -48,7 +48,7 @@ class IXloader(Interface):
         """
         pass
 
-    def datastore_before_update(self, resource_id, existing_info, new_headers):
+    def datastore_before_update(self, resource_id, existing_fields, new_headers):
         """ Called by the loader just before it is about to modify the
         DataStore table for a resource (truncate, drop+recreate, or create).
         It allows plugins to inspect the difference between the current
@@ -56,14 +56,25 @@ class IXloader(Interface):
         example to log an activity when columns are added, removed or
         renamed.
 
+        Both ``existing_fields`` and ``new_headers`` are lists of field
+        dicts that contain at least an ``id`` key, so a plugin can compute
+        the diff symmetrically::
+
+            old_ids = {f['id'] for f in existing_fields or []}
+            new_ids = {h['id'] for h in new_headers}
+            added = new_ids - old_ids
+            removed = old_ids - new_ids
+
         :param resource_id: the ID of the resource whose DataStore table is
             about to be updated.
         :type resource_id: string
 
-        :param existing_info: a mapping of ``{field_id: info_dict}`` built
-            from the existing DataStore table's Data Dictionary, or ``None``
-            if the DataStore table does not yet exist.
-        :type existing_info: dict or None
+        :param existing_fields: the current columns of the DataStore table
+            (the internal ``_id`` column is excluded), or ``None`` if the
+            DataStore table does not yet exist. Each dict contains at least
+            ``id`` and ``type`` and may include ``info`` for fields with a
+            Data Dictionary entry.
+        :type existing_fields: list of dicts or None
 
         :param new_headers: the list of field dicts that will be written to
             the DataStore. Each dict has at least an ``id`` and ``type``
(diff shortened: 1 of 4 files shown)


ckanext/xloader/loader.py
```diff
@@ -356,6 +369,7 @@ def load_csv(csv_filepath, resource_id, mimetype='text/csv', allow_type_guessing
         '''
         fields_match = _fields_match(fields, existing_fields, logger)
         if fields_match == FieldMatch.EXACT_MATCH:
+            _notify_datastore_before_update(resource_id, existing_info, fields)
```
ThrawnCA: Line 33 uses kwargs, this uses args. Should be consistent.

ThrawnCA (changes requested):
This looks good, thanks for adding it! 

Just need to make the calls to the new function use a consistent argument style.

ckanext/xloader/interfaces.py
```diff
@@ -47,3 +47,30 @@ def after_upload(self, context, resource_dict, dataset_dict):
...
+            if the DataStore table does not yet exist.
+        :type existing_info: dict or None
+
+        :param new_headers: the list of field dicts that will be written to
+            the DataStore. Each dict has at least an ``id`` and ``type``
+            key, and may include an ``info`` dict for fields that already
+            existed.
+        :type new_headers: list of dicts
+
+        The return value is ignored.
```
ThrawnCA, line 91: Should plugins have the opportunity to alter the new headers?

ckanext/xloader/loader.py
avdata99: Updated here https://github.com/okfn/ckanext-xloader/commit/8b2a7065b435f34141afca2e422ea2d38d5c4ed4 @ThrawnCA

ckanext/xloader/interfaces.py
avdata99: We don't need it but eventually someone could need it
I just add a warning:

> The ``existing_fields`` and ``new_headers`` lists are the
            same objects that the loader will use after this hook returns.
            Mutating them (e.g. adding, removing or renaming fields) will
            affect the subsequent DataStore operation.  This hook is
            intended for **read-only observation**; modify the lists only
            if you fully understand the downstream consequences.

What do you think?

ThrawnCA: Mm, I'm not sure we need to make the warning quite so dire? Just document the fact that fields can be altered and leave it at that, I reckon.

If someone is using this hook at all, they presumably have some understanding of the datastore field structure, and if they want to change things, okay.

avdata99: Updated here @ThrawnCA https://github.com/avdata99/ckanext-xloader/commit/588581cef98212ee0a35362c0b4926b8cad0c882
What do you think?

ckanext/xloader/tests/test_loader.py
```diff
@@ -1094,6 +1095,107 @@ def test_shapefile_zip_python3(self, Session):
...
+        )
+
+        with mock.patch.object(loader, '_notify_datastore_before_update') as notify:
+            loader.load_csv(
+                get_sample_filepath("simple2.csv"),
+                resource_id=resource_id,
+                mimetype="text/csv",
+                logger=logger,
+            )
+
```
ThrawnCA: Why two blank lines?

ckanext/xloader/interfaces.py
```diff
@@ -47,3 +47,50 @@ def after_upload(self, context, resource_dict, dataset_dict):
...
+
+        :param new_headers: the list of field dicts that will be written to
+            the DataStore. Each dict has at least an ``id`` and ``type``
+            key, and may include an ``info`` dict for fields that already
+            existed.
+        :type new_headers: list of dicts
+
+        .. warning::
+
+            The ``existing_fields`` and ``new_headers`` lists are the
```
ThrawnCA: I don't think the existing_fields object is the same after removing `_id`.

ckanext/xloader/tests/test_loader.py
avdata99: One line removed here
https://github.com/ckan/ckanext-xloader/pull/275/changes/b6a8b2e606c7774399cdcdfac7ebd2e6c456a9da

ckanext/xloader/interfaces.py
avdata99: Updated here https://github.com/ckan/ckanext-xloader/pull/275/changes/6e13330da5d924c5cc966d70388825d9f30a0855

ThrawnCA (approved):
@duttonw @wardi Thoughts?

wardi:
IIUC this is only useful for logging, but it looks fine to me.

ThrawnCA:
> IIUC this is only useful for logging, but it looks fine to me.

Theoretically a plugin could do things like overriding the data dictionary for specific fields when a resource is first created, instead of updating it later. Or eg allowing numeric fields while preventing any field from being treated as a timestamp.

duttonw:
Hi @avdata99 , 

Overall the logic you wish to add looks good and should also allow other plugins to 'block' table changes if the wrong file is uploaded with column's changed, i.e. structured data used in api/rest calls or join's and should not be altered. 

Do ensure your new unit tests to also be passing. (It may be best to use a different branch for quick iterations and then squash the commits then merge them onto this branch for cleaner history). 
It seems it was passing on the 16th but your code changes after that made it fail.  https://github.com/avdata99/ckanext-xloader/actions/workflows/test.yml?query=branch%3Adatastore_before_update++ 

https://github.com/avdata99/ckanext-xloader/actions/runs/24799367527/attempts/1#summary-72577605178 (4 tests failing)

Also, @wardi  it looks like we will all need to put some time into updating xloader as we now have 44 baseline tests having failed against latest ckan/master branch. :'(  
```============= 48 failed, 84 passed, 5 skipped in 105.31s (0:01:45) =============``` https://github.com/avdata99/ckanext-xloader/actions/runs/24799367527/attempts/1#summary-72577605148

avdata99:
Good catch @duttonw. Tests were updated here https://github.com/ckan/ckanext-xloader/pull/275/changes/94d7b68830da3b6fd8efd4182fe0570e55014804

Now they're green https://github.com/avdata99/ckanext-xloader/actions/runs/24833308497

duttonw (approved):

PR #280, opened 2026-06-23
CKAN v2.12 compatibility

smotornyuk:


duttonw (approved):
Can you squash commits to hide your debug testing. Otherwise it looks good

smotornyuk:
Yes, sure. Tests for v2.12 are currently red due to issues with the corresponding Docker-Postgres image. I reported the problem [here](https://github.com/ckan/ckan/issues/9364#issuecomment-4783781810). Tests on master are green, so I suspect that everything will be solved as soon as the Docker image is updated

amercader:
@smotornyuk I've updated the `ckan/ckan-postgres-dev:2.12` image but the tests seem to get stuck in `test_chuncks.py`. There must be something weird with the image, I'll investigate.
BTW `ckan/ckan-postgres-dev:master` doesn't get updated automatically, so it must be still on 2.11

smotornyuk:
@amercader, I identified the issue. The analyze query executed to estimate the row count in the datastore table was not explicitly closed, and sometimes it led to a deadlock with the following create-index query.

I rewrote the analyze-request to use a with-block so that the connection is properly terminated (similar to `with open(filename):...`). Now all test pipelines are green:

```diff
-   conn = engine.connect()
-   conn.execute(sa.text("ANALYZE \"{resource_id}\";"
+   with engine.connect() as conn:
+     conn.execute(sa.text("ANALYZE \"{resource_id}\";"
```

duttonw:
Seems we need at least one more test for overall code coverage. I can accept it now but it does highlight via codecov that this patch altered code not under unit (maybe under integration tho)

smotornyuk:
I don't want to be the one to blame for reduced code quality :smile: 

Added tests for reported lines

duttonw (approved):

PR #283, opened 2026-09-02
Make unicode headers configurable via ckanext.xloader.unicode_headers

tsila30:
encode_headers() previously always ran unidecode() on column headers, which transliterates or empties out non-Latin scripts (Hebrew, Arabic, Cyrillic, CJK). This adds a ckanext.xloader.unicode_headers boolean option (default False, preserving current behaviour). When True, headers are kept as-is (str) instead of being transliterated.

This modernises the approach from the OriHoch-add-support-for-unicode-headers branch, which relied on the Python 2 unicode() builtin and is broken on Python 3. Here we use p.toolkit.asbool(config.get(...)) and str.

Also declares the option in config_declaration.yaml and adds tests covering both the default transliteration behaviour and unicode preservation when enabled.

### DIFF
--- a/ckanext/xloader/loader.py
+++ b/ckanext/xloader/loader.py
@@ -690,7 +690,7 @@ def encode_headers(headers):
     # For non-Latin scripts (Hebrew, Arabic, Cyrillic, CJK, ...) unidecode
     # can mangle or empty out a header entirely, so setting
     # ckanext.xloader.unicode_headers = True keeps the original text as-is.
-    if p.toolkit.asbool(config.get('ckanext.xloader.unicode_headers', False)):
+    if p.toolkit.asbool(config.get('ckanext.xloader.unicode_headers')):
         decode_func = str
     else:
         decode_func = unidecode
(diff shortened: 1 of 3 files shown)


ckanext/xloader/loader.py
```diff
@@ -490,12 +490,21 @@ def get_types():
 
 
 def encode_headers(headers):
+    # By default, transliterate headers to ASCII with unidecode, so that
+    # e.g. accented Latin characters are simplified rather than dropped.
+    # For non-Latin scripts (Hebrew, Arabic, Cyrillic, CJK, ...) unidecode
+    # can mangle or empty out a header entirely, so setting
+    # ckanext.xloader.unicode_headers = True keeps the original text as-is.
+    if p.toolkit.asbool(config.get('ckanext.xloader.unicode_headers', False)):
```
duttonw: do be aware that since your 'config_declaration.yaml' has a 'default' value of false. you don't need to set the 2nd arg on this which can cause confusion when the yaml is updated and where its used is not changed (for defaults).

so please change this to
```
p.toolkit.asbool(config.get('ckanext.xloader.unicode_headers')):
```

tsila30: Done — rebased on master and removed the redundant default arg in config.get.

duttonw (approved):

PR #284, opened 2026-09-03
Convert to unidecode_headers and default to True

duttonw:
Re-reviewing https://github.com/ckan/ckanext-xloader/pull/283 

See against master for tweaks ontop of suggested feature flag: https://github.com/ckan/ckanext-xloader/compare/master...correct_behavour_for_patch 

Its better to have the feature flag mean what it does. 

Also add warning that if you don't 'convert' you can have issues. Use (as in not tested and may not be the most secure)

### DIFF
--- a/ckanext/xloader/config_declaration.yaml
+++ b/ckanext/xloader/config_declaration.yaml
@@ -71,15 +71,16 @@ groups:
 
             Strict means that a type will not be guessed if parsing fails for a single cell in the column.
         type: bool
-      - key: ckanext.xloader.unicode_headers
-        default: False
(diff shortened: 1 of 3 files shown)


tsila30:
Thanks for re-reviewing #283 and for the tweaks in correct_behavour_for_patch.

Renaming to `unidecode_headers` with a secure default of True makes sense: the flag now clearly describes what it does, and disabling conversion is an explicit, at-your-own-risk choice. The added warning about the security implications of not converting is a good call too.

All good on my side, 
do i have to open PR with the master ? how its work? 
@duttonw

PR #287, opened 2026-09-06
fix: drop trailing blank cells from wide rows in load_csv and load_table

### LINKED ISSUE #272: CSV loading fails when rows have trailing empty cells
Problem
When loading CSV files where some rows have extra empty cells at the end (common with Excel exports), XLoader fails because the row length doesn't match the number of headers/fields.

Error Example
The loader silently skips rows or fails when len(row) != len(fields) because of trailing empty cells.

Current Behavior
def strip_white_space_iter():
    for row in super_iter():
        if len(row) == len(fields):  # This condition fails with trailing empty cells
            for _index, _cell in enumerate(row):
                if fields[_index].get('strip_extra_white', True) and isinstance(_cell, str):
                    row[_index] = _cell.strip()
        yield row

The current code only processes rows when len(row) == len(fields), which means:

Rows with trailing empty cells are yielded without processing

This can cause data inconsistencies or loading failures

Expected Behavior
Trailing empty cells should be automatically removed before processing, so all valid data rows are loaded successfully.

Proposed Solution
Trim trailing empty cells before checking row length:

def strip_white_space_iter():
    for row in super_iter():
        # Remove trailing empty cells
        while len(row) > len(fields) and (not row[-1] or str(row[-1]).strip() == ''):
            row.pop()
        
        if len(row) == len(fields):
            for _index, _cell in enumerate(row):
                if fields[_index].get('strip_extra_white', True) and isinstance(_cell, str):
                    row[_index] = _cell.strip()
        yield row

Reproduction Steps:
Create a CSV file with headers: col1,col2,col3

Add data rows with trailing commas/empty cells: value1,value2,value3,,,

Try to load the file with XLoader

Observe loading issues

Additional Context
This is a common issue with:

CSV files exported from Microsoft Excel (adds extra empty columns)

Manually edited CSV files with inconsistent column counts

Data from automated systems that pad rows with empty values

The fix is minimal and handles the edge case gracefully without affecting normal CSV processing.

cgoldshtein:
## Summary

Excel and some other exporters pad every row with extra empty cells (trailing
commas). The header row ignores them, but data rows did not, so rows came out
wider than the declared columns and the DataStore `COPY` failed with
`extra data after last expected column`.

This makes data-row handling tolerant of surplus **trailing blank** cells,
matching how the header is already treated, while still rejecting surplus cells
that contain real data so we never silently drop it.

## Related

- Fixes #272 — CSV loading fails when rows have trailing empty cells.
- Relates to #274 — upstream PR addressing #272 (among other changes).
- Relates to qld-gov-au/ckanext-xloader#89 — handle empty cells past end of row.

## Changes

- Add `_classify_extra_cell(index, cell, header_count)`, a single place that
  decides how to treat a cell beyond the headers:
  - within the columns -> keep
  - surplus blank cell -> drop
  - surplus cell with real data -> raise `LoaderError`
- `load_csv`: `_make_whitespace_stripping_iter` becomes
  `_make_row_normalizing_iter`, which trims surplus trailing blank cells before
  the existing whitespace stripping so wide rows line up with the columns.
- `load_table`: reuse the same helper so both load paths behave identically.

## Behavior change

Rows whose surplus cells hold real data now fail fast with
`Found data in column N but resource only has M header(s)` instead of being
skipped or failing later. This surfaces genuinely malformed data early rather
than losing it silently.

## Tests

- `test_with_extra_blank_cells` / `test_with_extra_blank_cells_data_only` —
  rows padded with trailing empty cells load successfully via `load_csv`.
- Updated the KML unhandled-types assertions to expect the new
  `Found data in column ...` error, since a single-column CSV with real
  surplus data is now rejected up front rather than at field-definition
  validation.

### DIFF
--- a/ckanext/xloader/loader.py
+++ b/ckanext/xloader/loader.py
@@ -54,7 +54,7 @@ tabulator_config.CSV_SAMPLE_LINES = CSV_SAMPLE_LINES
 SINGLE_BYTE_ENCODING = 'cp1252'
 
 
-def _classify_extra_cell(index, cell, header_count):
+def _should_keep_cell(index, cell, header_count):
     """Decide what to do with a cell whose position may exceed the headers.
 
     Some exporters (notably Microsoft Excel) append extra empty cells to the
@@ -441,14 +441,14 @@ def load_csv(csv_filepath, resource_id, mimetype='text/csv', allow_type_guessing
         # (e.g. Excel exports that append empty trailing cells) still line up
         # with the declared columns instead of failing COPY with "extra data
         # after last expected column". A surplus cell holding real data raises
-        # a LoaderError via _classify_extra_cell rather than being dropped.
+        # a LoaderError via _should_keep_cell rather than being dropped.
         def normalize_row_iter():
             for row in super_iter():
-                # Trim trailing cells beyond the header. _classify_extra_cell
+                # Trim trailing cells beyond the header. _should_keep_cell
                 # returns False for a surplus blank (safe to drop) and raises
                 # if a surplus cell holds real data.
                 while len(row) > field_count and \
-                        not _classify_extra_cell(len(row) - 1, row[-1], field_count):
+                        not _should_keep_cell(len(row) - 1, row[-1], field_count):
                     row.pop()
 
                 if len(row) == field_count:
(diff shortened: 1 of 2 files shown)


ckanext/xloader/loader.py
```diff
@@ -54,6 +54,33 @@ def _notify_datastore_before_update(resource_id, existing_fields, new_headers):
 SINGLE_BYTE_ENCODING = 'cp1252'
 
 
+def _classify_extra_cell(index, cell, header_count):
```
ThrawnCA: Functionality looks good, but this name is not a good match for its behaviour. It's returning a boolean, not a classification, and the value is `True` if it's _not_ extra.

cgoldshtein: Renamed `_classify_extra_cell` → `_should_keep_cell` — reads more clearly
at the call sites (`if not _should_keep_cell(...)`). No behavior change.

duttonw (approved):

